PR #21, opened 2015-11-10
Fakeroot is needed for building a deb

michiil:
Add that fakeroot is needed for building a deb file. I had a problem with this so it would be good to add it.


EugeneKay:
I do not build for Debian, so I'll trust you on this.

PR #22, opened 2015-11-14
Fix file permissions on debian build with restrictive umasks

wummel:
Some more restrictive umask settings (ie. 0027) produce build errors when building the debian package. This patch properly sets all the permissions in the .deb package.


wummel:
fix a bug in the bug fix

EugeneKay:
Sorry, I can't accept changes to buildscripts.. The only changes I'm willing to merge must be non-conflicting with upstream(outside of licensing code). This basically limits to the INSTALL document. Please submit a patch to upstream(see subsonic.org).

Issue #42, opened 2016-05-05
Change major usages of "Subsonic"


EugeneKay:
The major usages of "Subsonic" should be changed to "Libresonic" in the user-facing portions of the application. This will reduce confusion associated with the Fork for new versions.

This issue is used to track finding and patching these changes. Please create a Pull Request and reference this Issue. If you have a comment, you can leave it here(refrain from commenting "Me too!", etc).

csobsidian:
As a note, I've started going through and finding these items. Obviously, they are quite numerous, but some are not as clear cut as they might seem. For instance, there are references to "Subsonic Premium" which have no context here. I think it would be wise to not touch any items that may be later slated for removal. That will make them easier to find. Thoughts?

EugeneKay:
I agree; let's leave "Subsonic Premium" references alone. Removing those(and squashing the remnants of the licensing code) are best left handled in a separate Issue further down the line.

Mostly, I want to make it clear that Libresonic is not the same software as Subsonic.

csobsidian:
Okay. I actually had some time today and am working through these now with the rebased code. A couple of more notes on things I am not changing at this time:

- Copyright notices that contain info about Subsonic. A new issue should be started (and dedicated) to tackling these.

- the getBrand() function of the SettingsService as I do not know how this will affect other operations.

csobsidian:
There are also a large number of references to subsonic website addresses in the localization files.

csobsidian:
Subsonic logos exist in the webapp for each theme (i.e. [logo.png](https://github.com/Libresonic/libresonic/blob/develop/subsonic-main/src/main/webapp/icons/buuftheme/logo.png)) and there is also a large format logo: [subsonic-512.png](https://github.com/Libresonic/libresonic/blob/develop/subsonic-main/src/main/resources/net/sourceforge/subsonic/service/subsonic-512.png).

csobsidian:
Do we want to put in a placeholder logo for these images until issue #43 is resolved and we have a new logo?

EugeneKay:
Merged #59.

Issue #43, opened 2016-05-05
Create a new Logo


EugeneKay:
A new logo is required for Libresonic. I do not have any art sklls, so community input is required. Please place your submission in the comments and I will arbitrarily pick the best one at an indeterminate future time. Please use GIthub's "reactions" emoji to vote; comments without a submission will be pruned.

williamkray:
![ls1](https://cloud.githubusercontent.com/assets/8409433/15064842/cb657560-130d-11e6-8c1d-463ffd5028df.png)

it's not amazing, but it's a start. Blue == freedom, L == libre, and soundwaves coming from a... speaker? cone? it'll get the ball rolling. trying to keep it clean, flat, and simple per current design trends.

williamkray:
![ls2](https://cloud.githubusercontent.com/assets/8409433/15065458/b4be0168-1314-11e6-8456-793a8c22b398.png)

variation with negative space. but i think this is getting too busy.

jolny:
I'm not a designer so I won't torment you with my paint skills. I just have a suggestion since I'm a bit punny. The german word "Lieber" means dear/beloved, etc. so maybe the new logo could contain a heart? Hearts also have round shapes, which would go well with soundwaves etc.
I like your second logo better @williamkray , I like the idea of blue and white together. Also, It's more "icon shaped" than the first one.

csobsidian:
I am proposing the use of the two following logos for the time being until this issue can be fully resolved.
![logo-large](https://cloud.githubusercontent.com/assets/8325573/15113675/58de20c2-15bb-11e6-9fc6-47b310d92b5d.png)
![logo-smallwide](https://cloud.githubusercontent.com/assets/8325573/15113676/58dfdaf2-15bb-11e6-924e-f27b2f0d49ae.png)

The small icon is used as a placeholder for the small format logo used by themes. Besides, who doesn't love scriptina?

jooola:
i Like William cray logo ! maybe merge it too ?

csobsidian:
Iterating on @williamkray's layout and color scheme. Let's keep this moving with ideas!
![logo-large](https://cloud.githubusercontent.com/assets/8325573/15147320/b87c808a-1685-11e6-83f0-580f76814521.png)
![logo-large2](https://cloud.githubusercontent.com/assets/8325573/15147324/bf28dc4e-1685-11e6-989f-74cf963f22c3.png)
[...]

PR #59, opened 2016-05-09
Change References from Subsonic to Libresonic

csobsidian:
This patch is meant to address issue #42.

This covers most of the string usage of "Subsonic", specifically those found in the default strings of the SystemService and those found in the localization files. I only changed non-English files where they seemed appropriate due to a change in the English localization.

This patch also changes the logos in all themes to a dark or light version of a placeholder image with the text "Libresonic". These images are not meant to be permanent. Rather, they are a stepping stone to keep moving forward with a new brand and remove as many user-facing references to Subsonic as possible.

No changes have been made to other references of Subsonic, such as those found in copyright notices, references to the net.sourceforge.subsonic namespace, or references to the subsonic website.


jooola:
But will they be any real logo ? ( something like a picture not only text ? )

EugeneKay:
Wow, thank you very much! This is way more involved than I was going to be, so that's awesome. Regarding the internationalizations, what is the significance of the (escape code changes)[https://github.com/Libresonic/libresonic/pull/59/commits/bcd38255b963863d4456c4628b7038a0ca77bc8e#diff-bfd7a120b400970aac44573605edf450L173]? I believe those are unicode sequences; is this to more closely follow the standards? I just want to ensure that we don't break other languages.

I am fine with that for a temporary logo. I welcome more entries and voting in #43.

Can you add a --signoff --allow-empty commit to the PR to acknowledge GPLv3(new as of 95f1ed7)? I will test and merge this evening.

csobsidian:
It looks like netbeans made all the escape code hex-characters lower-case: 00E1 becomes 00e1. I could not find a setting to leave them as they were.

EugeneKay:
Let's stick with the uppercase hex-codes. It would be best if you could rebase the bad change away, or I can do it before merging.

csobsidian:
I don't want to make extra work for you so I'll take care of it. I'm not very familiar with git or github so bear with me as I figure out the best way to go about it.

csobsidian:
Mildly off-topic but I'd like to make sure I do this right. (I'm happy to delete this comment later) Will this work:

- New branch from develop.
- Cherry-pick the two good commits.
- Make changes to the serversettings and localization files
- Push to issue-42-patch

Will that push the fixed files to this PR?

EugeneKay:
That would work correctly, yes.

Github PRs store the relationship between the two branches, meaning that any time you update issue-42-patch in your repo it will be updated here.

csobsidian:
Okay, all fixed.

EugeneKay:
That's fine, your acknowledgement here is good enough for me. Merge incoming.

PR #88, opened 2016-05-19
Fixed parent artifact name

Wicked0ne:
I'm unfamiliar with maven but libresonic was failing to compile when attempting to build a deb. A little digging and I found the artifactId in the pom files was incorrect. After making these changes it successfully compiled on debian 8.

### DIFF
--- a/subsonic-assembly/pom.xml
+++ b/subsonic-assembly/pom.xml
@@ -9,7 +9,7 @@
 
     <parent>
         <groupId>net.sourceforge.subsonic</groupId>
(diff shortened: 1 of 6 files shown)


chetan:
This was already fixed in the develop branch

Wicked0ne:
Doh! I guess I should have looked on here first. Oh well! I guess this should be closed now?

chetan:
Yup, I think so. I actually made the same mistake when I first checked it out too.

Issue #76, opened 2016-05-14
Remove or incorporate Subsonic Premium features


EugeneKay:
The Subsonic Premium feature set should be removed or incorporated(where possible).

EugeneKay:
This has been made part of the 6.2 Release Plan.

PR #93, opened 2016-05-21
Disable URL redirection

chetan:
This disables a public service provided by subsonic.org. Currently just short-circuits the relevant code and shows a warning in the UI and the logs that the feature is not available.

I vote to remove it completely but want to see what others think...

### DIFF
--- a/libresonic-main/pom.xml
+++ b/libresonic-main/pom.xml
@@ -279,6 +279,7 @@
             <scope>test</scope>
         </dependency>
 
+        <!-- UPNP / DLNA -->
         <dependency>
             <groupId>org.fourthline.cling</groupId>
             <artifactId>cling-core</artifactId>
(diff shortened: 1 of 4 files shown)


EugeneKay:
Per #76, this feature should either be removed completely or incorporated as a service that Libresonic provides. Currently it degrades gracefully(since the endpoint is not provided by libresonic.org), so this change should not be necessary? I am more comfortable leaving this feature as-is for now than having to revert this change later. We can review what would be necessary server-side to support this separately.

chetan:
Right. This mostly just cleans up the logs for me when developing. Otherwise it spews a bunch of errors/warnings. I think it's small enough to leave like this and decide later if there's demand for it. I personally don't need this feature.

afxgit:
To verify this just affects the *. subsonic.org service,  not the custom URL feature? I'm a user of custom URL for shares so hoping this wouldn't  affect that.

chetan:
@afxgit yes, this only affects the *.subsonic.org service, not the custom URL. That would still work as normal.

PR #94, opened 2016-05-22
Replace custom logger with slf4j/log4j2

chetan:
This wasn't really 100% necessary to clean up but I felt it was an easy starting point to get to know the codebase. Will get a bit cleaner once the move to spring 4.x happens. Planning to help on that branch next..


EugeneKay:
I have begun reviewing this patch, and I have a few questions:

- Why are there so many changes to the imports? It looks like they've been reordered by an IDE.
- What is the advantage of this logging library versus the existing one? I know that home-rolled isn't a great idea, but I'm not familiar with this library either.
- Will this cause any issues down-the-road with building against prepackaged targets?
- Why is "final" being added? Is this also an IDE thing? This seems unrelated to the rest of the change.

muff1nman:
I really appreciate the effort here and think this is a good change to make, but I don't think we can merge this as is at the moment:
- Lots of changes to not just logging, but also `@Override` and `final` changes. Not necessarily bad, but make it harder to do merges with changes that were done before this.
- Additional pom changes adding in jackson libraries, do we need them?
- I think we need to come up with a better way to deal with libresonic.home so we aren't setting it in multiple places. Really we probably need to integrate this with spring.

I see you put a lot of work into this and thank you for that, but I think changing the logging should wait until we are a bit more stable across the code base.

PR #95, opened 2016-05-22
Customizable From header

bajansen:
Customizable from header in advanced settings for password restore mail.

### DIFF
--- a/libresonic-main/src/main/resources/org/libresonic/player/i18n/ResourceBundle_nl.properties
+++ b/libresonic-main/src/main/resources/org/libresonic/player/i18n/ResourceBundle_nl.properties
@@ -343,8 +343,9 @@ advancedsettings.smtpPort = SMTP poort
 advancedsettings.smtpServer = SMTP server
 advancedsettings.smtpEncryption = SMTP versleutelingsmethode
 advancedsettings.smtpUser = SMTP gebruikersnaam
-advancedsettings.smtpPassword = SMTP wachtwoord
+advancedsettings.smtpPassword = Wachtwoord
 advancedsettings.smtpEncryption.none = Geen
+advancedsettings.smtpFrom = Afzender
 
 # personalsettings.jsp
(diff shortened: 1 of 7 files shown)


EugeneKay:
This looks great, thank you. I will test + merge later tonight.

libresonic-main/src/main/resources/org/libresonic/player/i18n/ResourceBundle_nl.properties
```diff
@@ -343,8 +343,9 @@ advancedsettings.smtpPort = SMTP poort
 advancedsettings.smtpServer = SMTP server
 advancedsettings.smtpEncryption = SMTP versleutelingsmethode
 advancedsettings.smtpUser = SMTP gebruikersnaam
-advancedsettings.smtpPassword = SMTP wachtwoord
+advancedsettings.smtpPassword = Wachtwoord
```
stevenroose: Why did you change this? It is the SMPT password that is asked here, right?

bajansen: Yup, but as I have rearranged the interface so the password field is now located behind the SMTP User field instead of underneath it, just putting 'Password' or in Dutch, 'Wachtwoord', looks cleaner than 'SMTP Password'. The password field for the LDAP Manager is also only titled 'Password'.

stevenroose: Is there a reason why this is only changed in the Dutch language?

bajansen: It's also changed in English, but the English language file also has some help text additions I haven't translated to Dutch so maybe that's why you missed it.

stevenroose: Aha, I see it. We should hook Libresonic up with a service like Transifex or something :p

PR #96, opened 2016-05-23
Install guide

alexfornuto:
I would prefer if someone would independently verify these instructions before it's merged.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -1,5 +1,5 @@
 <!--
-# INSTALL.md
+# BUILD.md
 # Libresonic/libresonic
 -->
 Requirements
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,0 +1,111 @@
+<!--
+# INSTALL.md
+# Libresonic/libresonic
+-->
+
+# Installing Libresonic
+
+This document is designed to explain how to install LibreSonic as a Tomcat module, on a computer running Linux. As the project expands, this guide will also expand to include other operating systems.
+
+## Installing From Pre-Compiled Package
+
+### Prerequisites
+
+In order to install and run Libresonic, you will need:
+
+  * A JDK installation. 1.8.x series of OpenJDK or Oracle JDK 8+ should work.
+  * A running [Tomcat](http://tomcat.apache.org/) server. If you're unfamiliar with Tomcat, there are many [guides](https://www.linode.com/docs/websites/frameworks/apache-tomcat-on-ubuntu-16-04) on it.
+
+
+### Download Libresonic
+
+Daily WAR files are built by Jenkins and available [here](https://jenkins.zifnab.net/job/libresonic/), curtesy of [zifnab06](https://github.com/zifnab06).
+
+1.  Download the latest war file:
+
+		wget https://jenkins.zifnab.net/job/libresonic/lastSuccessfulBuild/artifact/.repository/org/libresonic/player/libresonic-main/6.1.beta1/libresonic-main-6.1.beta1.war -O /var/lib/tomcat8/webapps/libresonic.war
+
+	Note that this command copies the war file directly to the Tomcat webapps directory, and renames it to `libresonic.war`.
+
+2.  Create the libresonic directory and assign ownership to the Tomcat system user (if running tomcat as a service):
+
+		mkdir /var/libresonic
+		chown tomcat8:tomcat8 /var/libresonic/
+
+3.  Start Tomcat, or restart it if running as a service, as in the example below using Systemd:
+
+		systemctl restart tomcat8.service
+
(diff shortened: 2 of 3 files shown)


cdetroye:
## BUILD.md
I have tested a build from source and deployment and the install worked fine.

**Platform**: Ubuntu 14.04.3 LTS
**JDK**: OpenJDK 1.7.0_101
**Maven**: 2.2.1
**Tomact**: 6

I know you said OpenJDK 8 in the build.md **but it was most certainly not needed**. Tomcat 6 also works. I have not yet tested it with maven 1. 

I am a bit confused why the deployment step for Tomcat is mentioned in the BUILD.md file, though. That is deployment instead of building and it is missing the extra step of creating the required directories so that might confuse readers. Tomcat throws some error about the database file being locked while the initial error is the directory not existing.

A minor note is that the clone of the current master is still branded as subsonic and not libresonic. That should match up as well. Maybe add a `git checkout develop`

m1dnight:
## BUILD.md
I have tested a build from source and deployment and the install worked fine.

**Platform**: Ubuntu 14.04.3 LTS
**JDK**: OpenJDK 1.7.0_101
**Maven**: 2.2.1
**Tomact**: 6

I know you said OpenJDK 8 in the build.md **but it was most certainly not needed**. Tomcat 6 also works. I have not yet tested it with maven 1. 

I am a bit confused why the deployment step for Tomcat is mentioned in the BUILD.md file, though. That is deployment instead of building and it is missing the extra step of creating the required directories so that might confuse readers. Tomcat throws some error about the database file being locked while the initial error is the directory not existing.

A minor note is that the clone of the current master is still branded as subsonic and not libresonic. That should match up as well. Maybe add a `git checkout develop`

edit: And finally, add that tomcat 6+ is a requirement.

alexfornuto:
Thanks for the review @m1dnight! I gave the versions that I was using on my build case. Having your input, I'll adjust the range of versions. 

The contents of BUILD.md remain largely unchanged for now. It's the former INSTALL file, which I am preserving until my INSTALL is set up, after which I will look at trimming BUILD.

Finally, I'll adjust all references to Subsonic.

uzantonomon:
I was reading the INSTALL.md file and the .WAR installation won't work if you create /var/libresonic. You have to create /var/subsonic. I'm using the Libresonic 6.0.1 and Tomcat 8.5.5.

a9e78rauijasdf:
Tried deploying on 16.04 on a clean install. Fails - from catalina.out:

[2017-03-06 17:53:57,114] WARN  org.springframework.web.context.support.XmlWebApplicationContext - Exception encountered during context initialization - cancelling refresh attempt: org.springframework.beans.fac
tory.BeanCreationException: Error creating bean with name 'liquibase' defined in ServletContext resource [/WEB-INF/applicationContext-db.xml]: Invocation of init method failed; nested exception is liquibase.exc
eption.DatabaseException: java.sql.SQLException: The database is already in use by another process: org.hsqldb.persist.NIOLockFile@497ae66d[file =/db/libresonic.lck, exists=false, locked=false, valid=false, fl
=null]: java.io.FileNotFoundException: /db/libresonic.lck (No such file or directory)
[2017-03-06 17:53:57,117] ERROR org.springframework.web.context.ContextLoader - Context initialization failed
org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'liquibase' defined in ServletContext resource [/WEB-INF/applicationContext-db.xml]: Invocation of init method failed; nest
ed exception is liquibase.exception.DatabaseException: java.sql.SQLException: The database is already in use by another process: org.hsqldb.persist.NIOLockFile@497ae66d[file =/db/libresonic.lck, exists=false, l
ocked=false, valid=false, fl =null]: java.io.FileNotFoundException: /db/libresonic.lck (No such file or directory)
        at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.initializeBean(AbstractAutowireCapableBeanFactory.java:1583)
        at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:553)
        at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:482)
        at org.springframework.beans.factory.support.AbstractBeanFactory$1.getObject(AbstractBeanFactory.java:306)
        at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.getSingleton(DefaultSingletonBeanRegistry.java:230)
        at org.springframework.beans.factory.support.AbstractBeanFactory.doGetBean(AbstractBeanFactory.java:302)
        at org.springframework.beans.factory.support.AbstractBeanFactory.getBean(AbstractBeanFactory.java:197)
 
[etc]

PR #104, opened 2016-06-11
Forum -> libresonic subreddit

bajansen:
Updated URLs to point to [https://www.reddit.com/r/libresonic](https://www.reddit.com/r/libresonic) instead of [http://forum.libresonic.org](http://forum.libresonic.org) as the latter doesn't exactly exist.


m1dnight:
I am not sure if this is up for discussion, but it might be nice to have an actual forum or Discourse instead of just a Reddit. 

It could be easily searchable for new users and could serve as a wiki for help.

Issue #101, opened 2016-06-06
Add keyboard shortcuts ?


jooola:
I'm up to add the keyboard shortcuts features to libresonic !
And easy to do ! commits are in the subsonic-svn repo !

Here is some more :
[https://github.com/Libresonic/subsonic-svn/commit/1213f8fc2a10655c0909eec631550131a10479dd](url)

[https://github.com/Libresonic/subsonic-svn/commit/12a0e1626461d98174d141e6920ead4e16268107](url)

Need to search in the subsonic-svn some other commits ( fixes and others adds for this features ).

I'll probably do it, but only when i have some time.

EugeneKay:
Personally I despise keyboard shortcuts in browser applications, but I understand that these are helpful. I am open to including this feature, however:

- It must be optional and off by default
- The commits referenced can not be cherry-picked directly, as they include unrelated changes
- Perhaps shortcuts should be user-customizable?

fxthomas:
Aren't the current keyboard shortcuts on by default? I, for one, have been using them all the time since Sindre first released 6.0-beta1.

I'm actually missing the "star" and "toggle queue" shortcuts present in 6.0-beta2, and I'd love to have them back in Libresonic!

afxgit:
Libresonic 6.0.1

First recommended release
Based upon Subsonic 5.3(stable)
v6.0 was tagged prematurely

fxthomas:
@afxgit Ha, thanks, I was out of country for a month and did not notice the rebase. Then let's just say that I'm also heavily interested in re-adding this feature to Libresonic! :smile: 

Call it scratching my own itch, but I will have some time to prepare a PR this weekend, is that okay with you or is somebody else already working on this?

Re: reusing code, anything wrong with just (for now) extracting the existing beta1 code and adding an optional _Enable Keyboard Shortcuts_ switch in under _Settings_ > _Personal_?

jooola:
As i sayd look at the different commits on subsonic-svn repo and write the code.
i didn't work on this features right now so you can peacefully start to work on it !

fxthomas:
@jooola Excellent, that's what I wanted to know :)

jooola:
Take note of the  EugeneKay comment !
- The commits referenced can not be cherry-picked directly, as they include unrelated changes !

fxthomas:
[...]

PR #110, opened 2016-06-18
Add keyboard shortcuts

fxthomas:
This PR is a work in progress towards adding keyboard shortcuts to Libresonic as discussed in #101. It's mostly a port of what was available in Subsonic 6.0-beta1, but I'd love to get feedback on how this is implemented.

Right now, the plan is to :
- [x] Add the same keyboard shortcuts as Subsonic (to avoid confusing users when they switch)
- [x] Test that JWPlayer behaves correctly
- [ ] Test that the Cast player behaves correctly
- [x] Test that the Jukebox player behaves correctly
- [x] Add a user option under the _Settings_ > _Personal_ page to enable/disable shortcuts
- [x] Add a help page that explains which keyboard shortcuts are available


fxthomas:
I've been meaning to ask this as well : is it okay to add a `togglePlayPause` method to both `CastPlayer` to the Jukebox REST API? I'm currently determining the play status by hand with the play/pause shortcut, which is at best unreliable, esp. for the Jukebox player.

fxthomas:
Work on this should be pretty much done, and I tested it on both the in-browser player and Jukebox mode.

The keyboard icon is the help page is pretty ugly, so if you are okay with bringing `font-awesome` in I can change it. Also shortcuts "should" work in ChromeCast mode (the code is there), but I don't have a device that supports casting so I did not test it at all.

EugeneKay:
Wow, this is huge. I will merge it into develop after 6.1 and we'll see how it goes.

fxthomas:
Perfect!

If we continue to follow the same conventions re: naming DB migrationn classes, they will need to be renamed for 6.2 once 6.1 is released. I'm currently running this on my server so I'll update the PR then.

fxthomas:
FYI : just rebased this onto `develop`.

plastikman:
Thank you for this @fxthomas I absolutely love them!

alexfornuto:
Where can someone see the shortcuts without going through the code itself?

fxthomas:
@alexfornuto Either press the '?' key or go to the "More" page (link on the top bar).

@plastikman Thanks :)

plastikman:
@alexfornuto I just noticed that in the documentation the keybinding icon to change songs in an arrow pointing up, the arrow should be pointing to the right.

http://imgur.com/a/tvFDx

fxthomas:
Hum, thanks for noticing! :)

plastikman:
in libresonic-main/src/main/webapp/WEB-INF/jsp/more.jsp
the code at line 219 needs to change from 
`<td class="more-shortcut">&#8593;</td><td class="more-shortcut-descr"><fmt:message key="more.keyboard.next"/></td>`
to 
`<td class="more-shortcut">&#8594;</td><td class="more-shortcut-descr"><fmt:message key="more.keyboard.next"/></td>`

Issue #122, opened 2016-08-10
Wrong version of servlet and jsp api


biconou:
Regarding the installation documentation, libresonic targets java8 and Tomcat8 (that's good).

But, in the libresonic-main pom.xml, the versions are wrong for servlet and jsp api.

```
    <dependency>
        <groupId>javax.servlet</groupId>
        <artifactId>servlet-api</artifactId>
        <version>2.4</version>
        <scope>provided</scope>
    </dependency>

    <dependency>
        <groupId>javax.servlet</groupId>
        <artifactId>jsp-api</artifactId>
        <version>2.0</version>
        <scope>provided</scope>
    </dependency>
```

Indeed Tomcat 8 uses these versions of the apis : 
Servlet 3.1
JSP 2.3

see : http://tomcat.apache.org/whichversion.html

biconou:
Ok, I am currently working on that issue. I also manage to add the support of jetty as  lightweight server for development purpose.

fxthomas:
Jetty support will be really useful, deploying to Tomcat during development is a pain! Does Libresonic have a developer "getting started" page I'm not aware of? If so it should be there.

biconou:
Good news. Jetty plugin support is now available on a PR I submitted. 
I planned to work on a guide to setup a development environment. 
But at the moment I am working on a port on Jetty in production mode as a standalone jar. There are some tricky stuffs.

biconou:
I just added support of standalone application with Jetty in my PR.
Enjoy.

muff1nman:
We are at 3.1 now, so this can be closed:
```
        <dependency>
            <groupId>javax.servlet</groupId>
            <artifactId>javax.servlet-api</artifactId>
            <version>3.1.0</version>
            <scope>provided</scope>
        </dependency>

```

PR #123, opened 2016-08-11
Fix for issue #122

biconou:
Fix for issue #122 

The libresonic/pom.xml has been cleaned to fit Tomcat 8 requirements. 
In addition, I have hadded the support of Jetty maven plugin thus it is very easy to start/debug the application inside IDE (if PR merged, I will document that).
To start libresonic in develoment mode : mvn jetty:run

Tests: 
Jetty tested successfully on theses plateforms
- Windows, Oracle JDK 8, Maven 3, IntelliJ 14
- Debian, Oracle JDK 8, Maven 3, IntelliJ 16

Build OK

Application libresonic.war deployed and tested successfully on plateform
- Windows, Oracle JDK 8, Tomcat 8
- Debian, Oracle JDK 8, Tomcat 8


biconou:
Oops, don't integrate this PR to soon. The latest commit broke the jetty-maven plugin. I'll fix these days.

biconou:
Well I removed the standalone packaging added in commit cf55abc. It was not a good idea as : 
- it breaks the jetty maven plugin
- it's useless at the moment as a standalone packaging already exists through the libresonic-assembly module.

muff1nman:
Awesome, definitely needed! However, I thought we had a dependency on apache taglibs in the custom url tag being used in the jsp pages. UrlSupport I believe is the name of the class.

biconou:
What is needed ? The standalone jar ?

fxthomas:
For me this only worked when first installing artifacts to the local repository with :

```
mvn install
```

...then running Jetty from the submodule :

```
mvn -pl libresonic-main jetty:run
```

Is that normal?

muff1nman:
Ah I see, that dependency is now resolved since you made the jstl an actual dependency.

muff1nman:
@fxthomas I think that is pretty normal. You can get away without doing a install only when executing within the context of the top level pom.

muff1nman:
The one thing that bothers me with this change is the fact that tomcat already provides most of these dependencies and hence why they were previously marked "provided". Perhaps it would be better to make this drive by a profile?

biconou:
That is an interesting objection. 
I am away this week. 
I'll investigate later.

fxthomas:
@muff1nman These commands were actually run from the root directory. Just running `mvn jetty:run` from there fails to find the Jetty plugin in the root module. Adding the plugin to the root POM does not run correctly because it tries to use its sources -- and it has none, `libresonic-main` has them!

I may have missed something obvious, I'm not that familiar with Maven. And I got it to work in the end, so that's okay with me -- just being able to use Jetty is a huge help when trying out things!

muff1nman:
@fxthomas Sorry if I implied that it should be run from the project root, because as you've found that won't work. Really the only way to run it as you initially described: a mvn install on the root and then running the plugin from the libresonic-main directory.

fxthomas:
Thanks, that's what I wanted to know. FWIW you can add that Jetty also works on OpenJDK 8 and Maven 3 then!

biconou:
@muff1nman I declared servlet and jsp api provided. It seems sufficient. Maven jetty plugin still works.

fxthomas:
I just found out that this PR breaks the bootstrap JAR (from `libresonic-booter`) that contains all Libresonic dependencies in the standalone installer. Libresonic seems to start, but accessing a page results in a "class not found" error related to the Jasper library.

biconou:
@fxthomas I think that the libresonic-booter system could  be replaced by a much more simpler packaging included an embedded jetty or tomcat server. I tried it (see above) but it does not work. 
I would be interested in working on that using SpringBoot wich seems to be a very inovative and smart way to package modern standalone web apps.

muff1nman:
I agree that it would be nice to move to spring boot which I have used many times in the past, but do we know how many people are using the libresonic-booter system?

biconou:
I think that the libresonic-booter system is heavy and confusing. Nowadays web applications are packaged as standalone jar or war to deploy  in a servlet container. libresonic is a mix between both: it is a jar running jetty and deploying a war. As it was previously recommended to run libresonic under Tomcat 8, I think that the libresonic-booter is useless and could be removed. 
I will probably work on a new packaging based on spring-boot let say in 2 month (cause i'm stuck on something else at the moment).

fxthomas:
@muff1nman I'd say a lot of people are using it, basically anyone who uses packaged releases (DEB, RPM, standalone, Windows and Mac installers).

@biconou Agreed. This confused me quite a bit back when I started using Subsonic. Glad to know you're working on it!

biconou:
I think that this PR should be abandoned. Everything here should be reconsidered with the SpringBoot migration goal.

muff1nman:
Agreed.

PR #126, opened 2016-10-20
Removed iml files

muff1nman:


biconou:
You also should exclude .iml and .idea from git source control. 
Just add these two lines at the end of .gitignore

.idea
*.iml

muff1nman:
Good idea. I have these in my global gitignore, but it would be good to have them in the project too.

EugeneKay:
This looks good to me.

Issue #58, opened 2016-05-09
Add Let's Encrypt support


IwishIcanFLighT:
Hi there,

I'm not gonna beat around the bush here: [privacy matters](https://blog.mozilla.org/blog/2016/02/16/help-us-spread-the-word-encryption-matters/).

After a long beta, [Let's Encrypt](https://letsencrypt.org/) was recently released but the current implementation of SSL in Subsonic doesn't make things easy.
[I've made myself a quick script to "facilitate" the process](http://forum.subsonic.org/forum/viewtopic.php?f=6&t=9400&p=70255#p70975), but there's definitely room for improvements imho.

Adding a seamless SSL implementation to Subsonic would encourage people to use SSL more frequently.

EugeneKay:
I completely agree: TLS/SSL is important, and should be used everywhere.

However, I do not believe that there is any point in implementing ACME(the Let's Encrypt protocol) at this time. Libresonic is primarily packaged as a WAR to be run inside of an application server(eg, Apache Tomcat), and does not directly speak HTTP by itself. The subsonic-booter application is used for the RPM/Deb/Exe targets and includes HTTP/S support, but I am not presently building against those targets and do not have any plans to do so.

I will leave this Issue open to solicit patches.

stevenroose:
I think the better option for supporting TLS is by using a reverse proxy like nginx.

jooola:
Could we think about Haproxy too ? i'm using it and don't get this working ...

IwishIcanFLighT:
My current setup already uses a reverse proxy with Apache.

The issue I'm facing is that Apache needs to bind a HTTPS virtual host to a HTTPS listening host.

Binding a reverse proxy for a HTTPS VH to a HTTP host such as LibreSonic on `http://0.0.0.0:4040` results to a handshake error.
Same goes for using a self-generated certificate on LibreSonic side.

Beside, I don't think having a non SSL host listening behind a SSL reverse proxy is a great idea in terms of security, but don't quote me on that.

In order to make it work properly with a behind proxy, you have to convert the certificate to LibreSonic's format so that both LibreSonic and Reverse Proxy use the "same" certificate.

That being said, I completely understand that it's not a priority at this point in time for LibreSonic, I just want to point that out for future releases. :wink:

csobsidian:
I'm positive that there is a way to get Apache to proxy your HTTPS into HTTP. There are many places that do this on their backend to reduce the server load from the encryption.

[...]

PR #129, opened 2016-10-23
Fix for Issue #58 - Add Let's Encrypt support

fuzzynectarine:
I'm not sure if this is the direction that you want to go, but this allows the libresonic-booter script to automatically create a keystore from a user defined base64 ssl certificate and key pair.


EugeneKay:
While encryption everywhere is a good goal, I don't believe that Libresonic should be handling SSL or tying itself to a single SSL vendor. That is a job for a reverse-proxy, eg Apache or Nginx.

EugeneKay:
Scratch my previous comment - I misread the patch. This looks like a sane patch at first glance, but needs to be looked-at in depth(as it is security-related code).

EugeneKay:
Libresonic-booter will be removed for the 6.2 release and replaced with Spring Boot.

Issue #71, opened 2016-05-12
Queue shuffle behaviour


EugeneKay:
The "shuffle" function for the Queue should move the currently-playing track to the top, instead of randomly within the list.

PR #132, opened 2016-10-28
Queue shuffle behavior

fxthomas:
I think this should fix the low-hanging fruit that is #71 : the currently-playing track is placed on top of the shuffled play queue.


EugeneKay:
You may be my new hero. As soon as I get my development instance back up(maybe this weekend) I'll give this a merge.

EugeneKay:
I just spent 5 minutes playing with the "Shuffle" button. I don't know how I lived before.

muff1nman:
Fyi looks like this is not in the develop branch. Working on a merge right now.

PR #139, opened 2016-11-12
Fix overflow issue with non-square cover art images

fxthomas:
This PR tries to prevent non-square cover art images from overflowing their container in thumbnail views. For me this happened with some quite old and odd albums from the early 90's ; it is definitely not frequent, but the UI is not very pretty or usable when it happens.

I starred a few albums to demonstrate this behavior. This is before :

screenshot from 2016-11-12 16-28-53

...and this is after the fix :

screenshot from 2016-11-12 16-27-35


EugeneKay:
This looks good, thanks. Merge incoming.

Issue #80, opened 2016-05-16
Migrate to Spring 4.2.6


muff1nman:
Libresonic is currently on some ancient 2.X version of Spring. According to this [doc](https://github.com/spring-projects/spring-framework/wiki/migrating-from-earlier-versions-of-the-spring-framework), 2.X isn't supported anymore.

I think this upgrade is needed for a couple reasons:
- Better support for integration testing
- Few developers use/know how to work with spring 2.X
- Spring 3 and 4 have a bunch of cool features that would be helpful for streamlining/cleaning up the codebase
- Addresses possible bugs/security issues

I have already started working on this and these are the problems I've run into so far:
- [X] Pom: Update spring version to 3.X.X as an intermediate
- [x] Pom: Update spring version to 4.2.6.RELEASE
- [X] Pom: Remove spring artifact and replace it with spring-webmvc
- [x] Controller: Migrate Controllers extending SimpleFormController to annotations
- [x] Controller: Update spring bean controller definitions (would like to use component-scan, but that's another effort I'm afraid
- [x] Controller: Remove url mapping in xml 
- [x] DB: Migrate queryForInt and friends to queryForObject
- [X] DB: Change usages of ParameterizedRowMapper to just RowMapper
- [ ] Other Unexpected Surprises
  - [x]  org.acegisecurity is not compatible with spring3. Rip it out and replace it with spring security.
    - [ ]  Update LDAP support to use spring security
    - [x]  Migrate to spring security remember me
    - [x] Fix login failure message

Spring 4 work is at: https://github.com/muff1nman/libresonic/tree/spring-4-update
Spring 3 work is at: https://github.com/muff1nman/libresonic/tree/spring-3-update

makkus:
Good idea. I'd wager this will be a fairly big change (it'd probably be also good to go through the codebase and look out for stuff that could be replaced by new spring features that didn't exist in Spring 2.X), but it'd be a great base to build further features upon.

EugeneKay:
Wow! This is an ambitious undertaking. I agree that  there are a lot of very old dependencies that should be updated, including Spring. I don't think this should negatively affect anything, but we'll need to test this a bunch before merging. I'm going to put this on the Future milestone for now; we'll need to test extensively before merging. It sounds like you know what you're doing; go for it!

I don't know what sort of scope / timetable this will take. Give me a shout if this needs a feature branch and I'll grant you some org-level permissions.

muff1nman:
[...]

PR #143, opened 2016-11-19
Spring 3 and Spring Security

muff1nman:
Although there is still work to be done to get to spring 4, I feel like spring 3 and migrating to spring security is a good milestone. See #80


biconou:
I tested this PR. It seems to work.
What must be done more ?

muff1nman:
Just needs review/testing and thanks for taking a look. There are a couple things to be done as listed in #80 but I think they can be done as we have time/see fit and shouldn't delay this any further. I think I wont merge this for a day or so to see if anyone else wants to review/test/merge it.

biconou:
I don't like the multiController (org.libresonic.player.controller.MultiController). It seems to be a messy multi-purpose controller thats serves multiple pages. What about splitting it in several specialized controllers ?

biconou:
What about moving the Spring configuration to annotation style ?

biconou:
In each xml Spring context file, the beans declaration still references the 2.5 version.

```
<beans xmlns="http://www.springframework.org/schema/beans"
       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
       xsi:schemaLocation="http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans-2.5.xsd">
```
It seems ok to replace 
http://www.springframework.org/schema/beans/spring-beans-2.5.xsd 
by
http://www.springframework.org/schema/beans/spring-beans.xsd

biconou:
May be there is some kind of problem with Spring Security. 
Here is the test case : 
Login Admin
I add a media folder (say music).
I scan the medias.
When I click on any artist name in the left frame, I got a message 

> Access denied
> Sorry, you are not authorized to perform the requested operation. 

Looking at the Admin user settings I can see that this user is granted to the music media folder.

muff1nman:
Would love to replace multi controller, just didn't get around to it.
Moving to annotation config is fine, I just didn't see it as terribly important.
Bumping the beans xsd is a good idea, that probably needs to be done.

PR #144, opened 2016-11-20
Second attempt at fixing the shuffle behavior

fxthomas:
This one was a bit hairier than I originally assumed : contrary to what I expected the `getCurrentFile` method never actually returned the currently playing file because... it was never updated when streaming.

This PR fixes that and the shuffle button now correctly floats the currently playing track to the top.


EugeneKay:
This has been cherry-picked into stable as 739fafe. Thanks!

Issue #151, opened 2016-11-27
Missing "Remember me" checkbox after Spring 3 migration


fxthomas:
The "Remember me" checkbox went missing after the migration to Spring 3. I use this, is there a way to get it back before the release?

Before → after :

![screenshot from 2016-11-27 19-00-20](https://cloud.githubusercontent.com/assets/613594/20650792/6ad72196-b4d7-11e6-98ab-7ddf960c6e33.png) ![screenshot from 2016-11-27 19-03-06](https://cloud.githubusercontent.com/assets/613594/20650794/6d13ad3a-b4d7-11e6-811e-42f97935ab8e.png)

muff1nman:
Yep, this one is my fault, I'll work on this.

jooola:
I think all the remember me features are not safe ! so maybe we could add a way to enable/disable this feature too ?

fxthomas:
@jooola Indeed they aren't. They also use MD5, so you definitely should not trust sensitive passwords to them.

I personally use them knowing this, and they are not dangerous if you don't check the checkbox, but I agree that not everybody will notice.

muff1nman:
@jooola This issue was mainly for fixing what previously worked before. As @fxthomas mentioned, you don't have to use the feature by not clicking the check box, but if you want to see a way to completely disable remember me, please open another issue and we can continue discussion there.

PR #154, opened 2016-11-28
Fix remember me

muff1nman:


muff1nman:
Addresses #151

muff1nman:
@fxthomas Can you test by any chance?

fxthomas:
I am afraid it doesn't work. I just checked the the `stable` branch to be sure : when I login with "remember me" then restart Libresonic, I stay logged in. This doesn't happen on `develop` even with this PR.

The docs mention adding a `RememberMeAuthenticationFilter` which supposedly handles these types of login requests, and this is where my knowledge of Spring stopped when I tried fixing it myself... Any other ideas?

muff1nman:
I had set it up so remember me keys refresh during application startup, but will revert it to its previous behavior.

muff1nman:
@fxthomas 409efd5 should get you the previous behavior.

fxthomas:
Awesome, it works for me!

PR #157, opened 2016-11-30
Fix SQL table and column being incorrectly swapped

TBastiani:
Fix SQL table and column being incorrectly swapped

The previous statement appears to incorrectly check for the column 'player' in the table 'm3u_bom_enabled'. Upon not finding such a column, it tries to add a 'm3u_bom_enabled' column into the 'player' table. On a fresh install this causes a SQL error since the column already exists.


muff1nman:
Yep, good catch, thanks!

EugeneKay:
@muff1nman This should be cherry-picked into develop as well.

muff1nman:
Done.

Issue #119, opened 2016-08-04
Documentation: Reverse-Proxying


EugeneKay:
Add documentation for adding a reverse proxy server(Apache or Nginx) in front of Libresonic under a Tomcat server.

fxthomas:
I've currently been using this for Nginx so that I can access Libresonic at `https://example.com/libresonic` :

```nginx
# Always redirect to the HTTPS version
server {
    listen      80;
    server_name example.com;
    return      301 https://$server_name$request_uri;
}

server {

    # SSL setup
    listen       443 default ssl;
    server_name  example.com;
    ssl_certificate      cert.pem;
    ssl_certificate_key  key.pem;

    # Libresonic Proxy
    location /libresonic {
      proxy_set_header X-Real-IP         $remote_addr;
      proxy_set_header X-Forwarded-For   $proxy_add_x_forwarded_for;
      proxy_set_header X-Forwarded-Proto https;
      proxy_set_header Host              $http_host;
      proxy_max_temp_file_size           0;
      proxy_pass                         http://127.0.0.1:4040;
      proxy_redirect                     http:// https://;
    }
}
```

Along with the following options in the startup script :

```bash
libresonic.host=127.0.0.1
libresonic.port=4040
libresonic.httpsPort=0
libresonic.contextPath=/subsonic
```

jooola:
Here is a configuration to use libresonic behind a reverse proxy like HAProxy : 


frontend http
	bind $server_public_ip$:80

	acl url_libresonic path_beg -i /libresonic
	use_backend libresonic-backend if url_libresonic

# Change default backend to libresonic backend if you don't have a web backend
	default_backend web-backend

backend libresonic-backend
	server subsonic 192.168.122.10:8443 check


# With ssl implementation

frontend https
	bind $server_public_ip$:443 ssl crt /etc/haproxy/ssl/$server_ssl_keys$.pem

	acl url_libresonic path_beg -i /libresonic
	use_backend libresonic-backend if url_libresonic

# Change default backend to libresonic backend if you don't have a web backend
	default_backend web-backend

backend libresonic-backend
	redirect scheme https if !{ ssl_fc }
	server subsonic 192.168.122.10:8443 check ssl verfiy none




Obsiously the libresonic context_path need to be changed.

nobswift:
ok, after struggeling for a week or so i found out how it works with nginx

   location ^~ /libresonic {
        proxy_set_header X-Forwarded-Host $host;
[...]

PR #165, opened 2016-12-06
Add documentation for reverse proxying

fxthomas:
This adds configuration snippets from #119 to a new documentation file on reverse proxying.

Additionally, shouldn't we clean up the docs and move `MIGRATE.md` and `BUILD.md` to the `documentation` directory so that they're in the same place? I can add this to this PR.


EugeneKay:
@fxthomas Yes, those should be moved & updated as well. I haven't validated the configs in your patch, but they look sane to me.

One addendum I would make is that if authentication is being performed at the proxy level(eg, using nginx's auth_request feature) then the Authorization header needs to be passed to the backend. Further, the /rest/ API endpoint does NOT support HTTP Basic auth, so it must be excluded from any such auth scheme. This is probably pretty specific to my use case(LDAP), so it may not make sense in the main docs.

fxthomas:
I added a commit for moving those files.

The configurations were directly taken from the parent issue in Apache and HAProxy's cases but sounded sane. The Nginx one is what I'm currently using, so I can confirm it works.

Why not add a specific section (in a subfolder) for "advanced" documentation topics? I'm thinking about writing some docs about the database and how one can tweak or retrieve things here, this could go there as well.

Issue #145, opened 2016-11-20
Continue random play queues after last song


fxthomas:
When using the "Shuffle" feature in the "More" section, the play queue stops at the last song. It should instead automatically add more random songs and start playing them.

While the whole playlist could be replaced with new songs, it would be nice if we kept one or two songs that were played before, so that the user can go back to them. For reference, this is what the [DSub](https://github.com/daneren2005/Subsonic/) app already does when shuffling songs.

The feature has already been [requested](http://forum.subsonic.org/forum/viewtopic.php?f=3&t=16373) on the main Subsonic forums. I have a very crude working version on [my local branch](https://github.com/fxthomas/libresonic/tree/local), so if there is interest in bringing it to Libresonic I can clean it up and make a PR.

EugeneKay:
I think that a "Shuffle Mode" is a good idea. I don't use/like the current implementations UI at all. I would welcome a Pull Request that replace it with a button to "Shuffle Library" and "Shuffle Starred". I think your local branch is a good starting point for this.

fxthomas:
Me neither. The current UI is hard to discover and not very intuitive. Also, in its current state there's no way for the user to tell whether the play queue is infinite.

On the other hand I use the filters quite a bit, e.g. to shuffle unstarred songs of a certain genre I haven't listened to for a while, so I'd like to keep them available even if they stay under "more".

I can think of two options :

* Make minimal options (shuffle library and shuffle starred?) easily accessible in the More dropdown next to the progress bar.
* In the future, "smart playlists" under the Playlists menu would be a great way to replace this with a more durable solution.

What do you think?

fxthomas:
I'm suggesting the name "shuffle radio" for this feature, anybody has a better idea?

fxthomas:
I'm not entirely satisfied with the UI, but this should be ready as far as features go. Anyone cares to test this?

Start a radio with the "shuffle radio" button under the "more" page :

![screenshot from 2016-12-11 15-20-54](https://cloud.githubusercontent.com/assets/613594/21080685/f1706030-bfb5-11e6-957c-c33938cca322.png)

New songs will be added to the play queue when you reach the end (replacing the "repeat all" feature). Stop the shuffle radio (but keep the previously played songs) at anytime with the "stop shuffle radio" link :

[...]

PR #168, opened 2016-12-07
Shuffle radio

fxthomas:
This PR is an implementation of what's described in #145.

- [x] Make random playlists reload songs when reaching the end
- [x] Add a visual indicator in the UI so that the user knows this is an infinite playlist
- [x] Add songs to the end instead of replacing
- [x] Let the user choose to enable/disable infinite in the shuffle section


zifnab06:
LGTM

PR #172, opened 2016-12-10
External Database Support

muff1nman:
### Changes
- Adds commons-configuration for more control over properties
- Allow ${} interpolation in spring config
- Uses liquibase abstraction for performing schema modifications
- Adds database configuration for three modes:
  - `legacy`
  - `jndi`
  - `embed`

See documentation/DATABASE.md for more details.

### Other Notes
- Please backup your database before reviewing this code
- There are some non-show-stopper bugs to be aware of:
  - https://issues.apache.org/jira/browse/CONFIGURATION-627
     - workaround in place
  - https://issues.apache.org/jira/browse/CONFIGURATION-644
    - properties file will not have header comment until this is fixed
  - https://liquibase.jira.com/browse/CORE-2944
    - workaround by using older version for now
  - https://liquibase.jira.com/browse/CORE-2966
    - see [comment below](https://github.com/Libresonic/libresonic/pull/172#issuecomment-269031087)

### TODO
- I need to do some more testing myself, but could use some help
- Add automated testing for our supported databases. Do we want to discuss what should be in this list?
- Do we want to provide users with builds of libresonic with the proper jdbc drivers?

### Dev notes
- Please do not modify liquibase scripts once they have been pushed to users. In the case we need to do this, we will need to add a list of previous liquibase changelog checksums so that we don't break users setups.
- Careful with keywords across all database vendors.


muff1nman:
Well looks like travis isn't happy.

muff1nman:
Currently running into issues with postgres due to the fact how insert statements are executed. Hsql is so backwards in some things...

fxthomas:
At work we use PostgreSQL in a linked Docker container for running unit tests on a real server. It works pretty well, maybe we can use something similar for Libresonic on Jenkins?

zifnab06:
First: "Oh god, they rolled their own SQL library". 

Second: if subsonic only runs in tomcat (which, to my understanding is accurate), IMO this should be moved over to use a JDBC connection w/ a Tomcat Resource.

muff1nman:
@zifnab06 it was my preference to use tomcat for the jdbc connection, but as discussed in the original issue, we might not have tomcat with spring boot. So I made it configurable between the two.

muff1nman:
Postgres and old hsql driver seem to be working okay with that last change.

muff1nman:
I don't think we need to wait for the TODO items to merge this, so this is ready for merging unless someone else wants to review and provide feedback.

fxthomas:
If you can wait a few days until this week-end I'd like to check that things work well on my end, especially since I don't use Tomcat for running Libresonic.

fxthomas:
Okay, after some testing, Liquidbase craps itself migrating my current HSQLDB database using either the `embed` and `legacy` methods. I did a `mvn clean` before compiling just to be sure.

This [gist](https://gist.github.com/fxthomas/55a3b8321466e56a399027377ec1682f) contains a log of Libresonic starting up, but here is a summary :

* When creating an index in migration 47: *Table not found: TABLE_CONSTRAINTS in statement*
* Warnings from Liquidbase about XML parsing: *Current XML parsers seems to not support EntityResolver2. External entities won't be correctly loaded*
* An `IntrospectionException` when accessing a property in the Booter code

One question: Liquidbase seems to be rolling back migrations that don't exist before actually starting again at migration 25, is that normal?

muff1nman:
Thanks for taking a look.

The IntrospectionException is caused by https://issues.apache.org/jira/browse/CONFIGURATION-627, but I had put an entry in the logger to mute it for now. So its interesting that the logger isn't taking effect for you.

I have not seen the "Current XML parsers seems to not support EntityResolver2. External entities won't be correctly loaded" before. It doesn't seem to impact the execution it seems.

The initial rollback creates a rollback.sql file to undo any liquibase changes. Not sure if we necessarily want this as it can be turned off.

The creating index issue is interesting, I'll need to look into that.

fxthomas:
This is a wild guess, but the two warnings (XML parser and logger configuration) are probably caused by the booter JAR, so they may go away once we migrate to Spring Boot.

Regarding the index issue, my database files are quite old, dating from Subsonic 5.2 if I remember correctly. I can provide an anonymized version if necessary.

muff1nman:
That would be awesome actually as I don't have an old version.

muff1nman:
Traivs, you're a lier, that last build should have failed: https://gist.github.com/muff1nman/3f3a19157baa023d6bd5a43a1131279d

muff1nman:
I created a defect with liquibase to address the index issue: https://liquibase.jira.com/browse/CORE-2966.

However we don't need to wait for them to fix it, as I have put a workaround in place in the meantime. Namely, we can move index creation into the same area the table is created. The downside is that folks that have a hsql database between the schema version the table was created and the schema where the index is created will be missing the index. (i.e. between schema 32 and schema 53). I am anticipating that there will be very few people affected by this and even if they are, I don't think these indices are critical. 

@fxthomas , mind taking another shot? Thanks.

fxthomas:
Appears to be working well. I tested various pages before and after a library scan, nothing to report. Great job!

Can't we use some raw SQL to insert those indexes, or do it in some custom migration class (I don't really know Liquidbase)? I don't care if the installation instructions say "upgrade to 5.3 first", but it would make it significantly harder for newcomers to upgrade.

muff1nman:
This should be a better workaround that doesn't miss indexes depending on how old people's databases are. Let me know what you think.

muff1nman:
Anyone have any other reservations/concerns about merging this?

PR #173, opened 2016-12-10
Store all Google fonts locally

fxthomas:
I'm proposing that we replace all external Google fonts by a locally-stored version. I can think of three reasons why this is a good idea :

* Libresonic is often used on local networks and should not depend on internet connectivity.
* Censorship (China comes to mind) and restrictive firewalls can break Google Font APIs.
* Privacy issues with using Google services.

I've downloaded (using [Google Webfonts Helper](https://google-webfonts-helper.herokuapp.com/fonts)) the two fonts (Ubuntu and Roboto) and made this PR, but what do you think?

I have two questions if this is accepted :
* I'm fairly sure that the Ubuntu/Roboto license should allow us to do this, but do we have to copy the license terms somewhere when including those files?
* I've included *all* formats here, but we should be able to use only WOFF and WOFF2 if we accept that [older browsers](http://caniuse.com/#feat=woff) such as Android < 4.4 and IE < 9 will fall back to the default sans-serif.


zifnab06:
Ubuntu Font License: http://font.ubuntu.com/ufl/ubuntu-font-licence-1.0.txt
Roboto Font License is Apache 2.0, which also requires attribution for redistribution.

EugeneKay:
Is Ubuntu Font License GPLv3-compatible?

Roboto should be fine; Apache is compatible with being distributed as GPL. The License section of README.md should be amended to state `Includes Roboto Fonts by Google, distributed under GPLv3. Available as Apache 2.0 from <url>`.

zifnab06:
I'm going to go ahead and close this, as the fonts aren't gplv3 compatible - I realize its a concern getting fonts from the internet for non-internet enabled clients, but I'd recommend adding a fallback font in css for that instead.

fxthomas:
Fine with me, thanks for having looked into this anyway :)

PR #174, opened 2016-12-12
Auto locate transcoder binary

muff1nman:


fxthomas:
Appears to be working for me on Linux!

muff1nman:
This was reverted as I found some issues with it that were not easily solvable. Mainly, the check for if a transcoder is available does not work with the autolocate feature.

PR #177, opened 2016-12-14
Migrate random play queues to Spring 4

fxthomas:
Some controllers were skipped during the Spring 4 migration. This is one I'm familiar with, so here is a PR.


biconou:
Indeed, I migrated controllers only when needed due to compilation problems. This strategy minimize regression risks. But it's a good idea to migrate them continiously when the occasion to test properly is present.

PR #184, opened 2016-12-19
Update StringUtil.java

Tak-MK:
I have literally no idea about Java, but I suppose my change will show TB data instead of GB when having more than 1024GB of music (I don't like seeing the "1120GB", it's weird ¯\_(ツ)_/¯).

Thanks!


stevenroose:
While you're at it, you could immediately update them to either use correct amounts for MB, GB and TB (10^6, 9, 12) or use the [binary prefixed](https://en.wikipedia.org/wiki/Binary_prefix) MiB, GiB and TiB.

zifnab06:
Hi, 

It looks like you broke a test case while doing this: 

https://github.com/Libresonic/libresonic/blob/develop/libresonic-main/src/test/java/org/libresonic/player/util/StringUtilTestCase.java#L62

Quick explanation: MAX_INT is the maximum number an integer can store - which is 2^32-1. 1024*1024*1024*1024 is calculated as an integer value, which is more than 2^32-1. 

I'm not actually sure what the correct format for java is (either 1024.0*1024.0*1024.0*1024.0, or use the BigInteger class), but I can't merge it if the tests are failing (as this would actually break all byte counts, not add TB). 

Thanks!

zifnab06:
Aaand, I'm an idiot. Longs are a thing (double width integers, 2^64-1). Use 1024l * 1024l * 1024l * 1024l.

Tak-MK:
Hum... sorry guys, maybe I should send it as a issue-idea, because I don't know how to program in Java :( (I thought it was easier)
Maybe I should close this PR?

alexfornuto:
@Tak-MK I'm not a programmer either, but I say you leave this PR open. It has all the information as an issue would, and may help someone who does know how to do this, by providing a starting point to work from.

muff1nman:
Added some tweaks and merged. Thanks for the PR.

PR #196, opened 2016-12-30
Use original file type for formatting

### LINKED ISSUE #131: FLAC files appear as MP3 in play queue
A small cosmetic thing is bugging me : FLAC files appear as MP3 in the play queue :

screenshot from 2016-10-28 22-25-00

Apparently `PlayQueueService.formatFormat` tries to get the format from the transcoder (mine always transcodes to MP3), which is somewhat misleading as bitrate column next to it still reports the bitrate of the original file.

Shouldn't Libresonic show the actual file format there to stay consistent?

DavidARussell:
The bitrate column uses the original bitrate of the file on the server,
so the 'format' column should use the original filetype rather than
whatever the file is transcoded to (to avoid confusing mismatches like
an '800kbps' MP3 file when transcoding from FLAC to MP3).

Fixes #131

Signed-off-by: David Russell <david.russell.scotland@gmail.com>


DavidARussell:
Closing as I filed this against the wrong branch, fixing.

PR #197, opened 2016-12-30
Use original file type for display formatting

### LINKED ISSUE #131: FLAC files appear as MP3 in play queue

DavidARussell:
The bitrate column uses the original file, so the file type column
should do the same for consistency.

Fixes #131

Signed-off-by: David Russell <david.russell.scotland@gmail.com>


DavidARussell:
Has the same code change as #196, but against the correct branch this time

PR #210, opened 2017-01-05
Restore PasswordSettingsController

biconou:
I previously removed the PasswordSettingsController but that was a mistake as it is need for a non admin user to change his password.

### DIFF
--- a/libresonic-main/src/main/java/org/libresonic/player/controller/PasswordSettingsController.java
+++ b/libresonic-main/src/main/java/org/libresonic/player/controller/PasswordSettingsController.java
@@ -1,0 +1,89 @@
+/*
+ This file is part of Libresonic.
+
+ Libresonic is free software: you can redistribute it and/or modify
+ it under the terms of the GNU General Public License as published by
+ the Free Software Foundation, either version 3 of the License, or
+ (at your option) any later version.
+
+ Libresonic is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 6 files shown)


libresonic-main/src/main/java/org/libresonic/player/controller/PasswordSettingsController.java
```diff
@@ -75,14 +67,17 @@ protected ModelAndView displayForm(HttpServletRequest request) throws Exception
...
+            command.setPassword(null);
+            command.setConfirmPassword(null);
+            request.setAttribute("settings_toast",true);
 
-        command.setPassword(null);
-        command.setConfirmPassword(null);
-        command.setToast(true);
+        }
         return "passwordSettings";
     }
```
muff1nman: We should probably do this similar to the other controllers and do "redirect:passwordSettings" and then set the "settings_toast" on the flash attributes.

libresonic-main/src/main/java/org/libresonic/player/controller/PasswordSettingsController.java
```diff
@@ -43,7 +43,7 @@
  */
 @org.springframework.stereotype.Controller
 @RequestMapping("/passwordSettings")
-@SessionAttributes( value="command", types={PasswordSettingsCommand.class} )
+//@SessionAttributes( value="command", types={PasswordSettingsCommand.class} )
```
muff1nman: We could just remove this line entirely.

biconou: That's fixed. Except that in case of error, I don't redirect.
Sorry, I'm a bit bored by Spring MVC.

PR #213, opened 2017-01-06
WIP : Modify assembly process

biconou:
This PR modifies the assembly process thus 
- it creates a standalone package containing the shell scripts and the spring boot war
- it creates a package containing the war to deploy to tomcat

The shell scripts have been moved from booter subproject to assembly subproject. 

In addition, the libresonic.sh shell adds an option to start JMX.


libresonic-assembly/src/main/assembly/war.xml
```diff
@@ -16,12 +16,12 @@
...
-            <outputDirectory></outputDirectory>
-            <includes>
-                <include>*.war</include>
-            </includes>
-        </fileSet>
     </fileSets>
+    <files>
+        <file>
+            <source>../libresonic-main/target/libresonic.war.original</source>
+            <outputDirectory></outputDirectory>
```
muff1nman: Why use libresonic.war.original over libresonic.war?

biconou: The .war.original package is the original .war package backed up by the spring boot plugin before repackage to standalone war. It is a "classic" war intended to be deployed to Tomcat.
war.xml is the assembly profile to produce the package to be deployed to tomcat while standalone.xml is the profile to produce the standalone package (ie containing the SpringBoot application).

biconou:
Travis got into anger because of the debian packaging.

libresonic-assembly/src/main/assembly/war.xml
muff1nman: Actually the libresonic.war should be fine to use as well. Although it contains some extraneous spring boot jars and configurations, spring boot puts them in the war such that they should not conflict with tomcat war deployment. I would rather we only have people use the libresonic.war to be consistent. In fact, perhaps we should even go so far as to not retain the libresonic.war.original file during the maven build process so we ensure people don't use it.

biconou: If the libresonic.war also works deployed on Tomcat that's ok. 
I'll fix it.

muff1nman:
Looks like we will need to update the debian package to not use the booter anymore.

libresonic-assembly/src/main/script/libresonic.sh
```diff
@@ -111,6 +119,7 @@ if [ -L $0 ] && ([ -e /bin/readlink ] || [ -e /usr/bin/readlink ]); then
 fi
 
 ${JAVA} -Xmx${LIBRESONIC_MAX_MEMORY}m \
+  ${JMX_OPTIONS} \
```
fxthomas: Just a thing : `contextPath`, `host` and `port` were not handled by Libresonic directly anymore last I checked. I had to add the Spring Boot equivalents I found in the docs to make it work properly :

```bash
-Dserver.host=${LIBRESONIC_HOST} \
-Dserver.port=${LIBRESONIC_PORT} \
-Dserver.contextPath=${LIBRESONIC_CONTEXT_PATH} \
```

muff1nman:
@biconou I don't think we need this anymore.

Issue #211, opened 2017-01-06
Database configurations broken


muff1nman:
Looks like the spring profiles aren't getting setup correctly, will investigate.

PR #215, opened 2017-01-11
Fix #212, activate legacy datasource profile

### LINKED ISSUE #212: Hsql database not getting saved?
Seems like the hsql database is not getting saved correctly. I've seen some cases where data seems to get lost. We might need to add this back in:

https://github.com/Libresonic/libresonic/blob/stable/libresonic-main/src/main/java/org/libresonic/player/dao/HsqlDaoHelper.java#L83

Japi42:
This fixed the database not saving issue in Libresonic for me (Issue #212).  It seems that the Spring wasn't pulling in the datasource profile so it wasn't actually creating a Hsqldb in-file database (must have being doing in-memory since everything worked, but /var/libresonic/db was empy).  This sets the active profile to "legacy" to enable the normal datasource (via DatasourceProfileActivatorPropertySource).  I'm not sure if this should actually be set somewhere else or if this is the completely wrong fix (maybe the embed and jndi profiles should be active as well?).  I haven't done Spring in many years so it was a bit of a struggle to figure out what was breaking and how to get my new install to actually save to disk.  This is also my first commit/pull request in git, sorry for any mistakes I am making!

### DIFF
--- a/libresonic-main/src/main/resources/application.properties
+++ b/libresonic-main/src/main/resources/application.properties
@@ -1,3 +1,4 @@
 spring.mvc.view.prefix: /WEB-INF/jsp/
 spring.mvc.view.suffix: .jsp
 server.error.includeStacktrace: ALWAYS
+spring.profiles.active: legacy


muff1nman:
Awesome! Ill take a look at these commits, but thanks for taking a shot at this!

muff1nman:
Interesting, so you ended up activating the profile manually. What was happening before the spring boot merge is that `org/libresonic/player/spring/DatasourceProfileActivatorPropertySource.java` was supposed to set it up for you based on your selection of `database.config.type`. It was a bit of a hack, but the idea was to not expose the `spring.profiles.active` to the user so we could avoid having spring-level concepts in folks `libresonic.properties` file. With this change, people would then have to set `spring.profiles.active` in their `libresonic.properties` to override the default `spring.profiles.active=legacy`. I'd like to avoid that if possible, so I'll need to research a bit on why you had to set this in the first place. Thanks again for digging into this!

muff1nman:
@Japi42 do you have some details on how you ran across this? Tomcat? Linux? etc? I am having a hard time reproducing the original issue now.

Japi42:
This was in Tomcat in Linux (Ubuntu), and I looked at it again based on how it is supposed to work and it is working fine now without my change.  I think what happened was I did my initial checkout over the weekend and I must have grabbed a bad version.  I banged against that until I got something that worked (the fix I submitted) but during that time the develop branch was fixed and the change isn't needed.  I suspect it was the change in #211 that resolved the issue correctly, but I hadn't updated to see that change until I forked/merged this evening.  That's what I get for not updating frequently, and doing a complete recreation of the problem in the base before committing.  So I spent 3 days trying to fix a bug that was already fixed, but at least I know a little more about Spring.  Sorry about the unneeded patch!

muff1nman:
Okay that would explain things. I'll close this one out but thanks for helping test the develop branch :)

PR #223, opened 2017-01-15
Remove installation from project

muff1nman:
Removes the booter, assembly, and installation submodules. There may still be some value in these deleted files, but it is expected that any packages will be moved to a different repository under the libresonic org.

In addition, I have removed the ability to change the port and context path as these should now be done either via tomcat or via spring boot. It may make sense to present options to change these like how we did in libresonic.sh, but again it should be done via spring boot at that point and not done in the main codebase.


biconou:
I think the libresonic-assembly is still usefull. It contains the shells.

muff1nman:
The shell scripts themselves might be useful, but libresonic-assembly itself is useless as the only artifact needed is the libresonic.war built in libresonic-main. At any rate, we could salvage the shells out to a different repo along with the other os-specific things later.

biconou:
Indeed. Since these other repos are not yet created, should'nt we store the shells anywhere else (for example in the documentation).
What about the libresonic-tools subproject ? It seams to be outdated too.

muff1nman:
I believe I did remove the tools directory. We can always go back in the history to look at the shell scripts.

EugeneKay:
@biconou @muff1nman Let me know if you'd like a separate libresonic-tools repo

biconou:
That could be usefull. Nice proposition.

PR #226, opened 2017-01-17
Tweak Tomcat performance

### LINKED ISSUE #221: Potential performance issue
In a previous discussion, @fxthomas mentionned that running Libresonic 6.2 as a standalone app could be slower than when it is deployed to Tomcat.
I would like to investigate that.

@fxthomas could you tell me exactly on what special usecase you noticed this problem ?

fxthomas:
This commit fixes #221 by doing the following things :

* Increase Tomcat resource cache size to be able to actually store
  something in there.
* Increase Tomcat resource cache TTL to avoid costly evictions
* Put Jasper in production mode to avoid recompiling JSPs


muff1nman:
lgtm

PR #247, opened 2017-02-11
Fix status page crash

fxthomas:
The crash occurs because the corresponding JSP template tries to access an inner class recently changed from `public` to `private`.


cryptix:
Came here to open an issue about this and found a fix. Nice!

Tested it and it's working again!

Issue #242, opened 2017-02-01
Haproxy reverse proxy configuration don't work for SSL


jooola:
So i tested the Haproxy configuration in the documentation
```
frontend https
    bind $server_public_ip$:443 ssl crt /etc/haproxy/ssl/$server_ssl_keys$.pem

    # Let Libresonic handle all requests under /libresonic
    acl url_libresonic path_beg -i /libresonic
    use_backend libresonic-backend if url_libresonic

    # Change default backend to libresonic backend if you don't have a web backend
    default_backend web-backend

backend libresonic-backend
  server libresonic 127.0.0.1:8080 check
```

And as expected nothing works !
If i use the current settings the server works but same bug as subsonic we have a white home page after login, it is simply empty.
Screenshot : 
![capture d ecran de 2017-02-01 20 08 01](https://cloud.githubusercontent.com/assets/19195485/22522041/64b84b5c-e8ba-11e6-9ddf-fb6eb56d48f4.png)
![capture d ecran de 2017-02-01 20 09 31](https://cloud.githubusercontent.com/assets/19195485/22522043/64c4a06e-e8ba-11e6-8f70-55ef1e13ded8.png)

**This is maybe due to the mixed content of the page not loading.**

With the configuration i wrote last time : 
```
frontend https
    bind $server_public_ip$:443 ssl crt /etc/haproxy/ssl/$server_ssl_keys$.pem

    acl url_libresonic path_beg -i /libresonic
    use_backend libresonic-backend if url_libresonic

    default_backend web-backend

backend libresonic-backend
    redirect scheme https if !{ ssl_fc }
    server libresonic 192.168.122.10:8443 check ssl verfiy none
```
I changed this since we don't have any ssl in the backend
`server libresonic 192.168.122.10:8080 check ssl verfiy none`

And now i have a 503 error, i guess because haproxy is expecting SSL communication but they aren't !

Maybe need to find a working haproxy configuration or integrate a patch to libresonic to fix this.
( simply remove the mixed content problem )
Here is an old patch from subsonic we tryed to add here but i solved the problem with a work around ( not working anymore )
#120 
https://sourceforge.net/p/subsonic/patches/23/

Hope this is enough detailed !

Btw need to change 
```
backend libresonic-backend
  server libresonic 127.0.0.1:4040 check
```
into
```
backend libresonic-backend
  server libresonic 127.0.0.1:8080 check
```
in the current documentation

fxthomas:
[...]

PR #260, opened 2017-02-26
Remove url redirection

muff1nman:
- Add upnp specific base url
- Use http request base url for areas used by url redirection
- Fix sharing
- Removed Network Settings tab
- Remove port forwarding


muff1nman:
Can someone review please?

libresonic-main/src/main/java/org/libresonic/player/service/NetworkService.java
```diff
@@ -237,4 +75,21 @@ public Date getDate() {
...
         }
     }
+
+    public static String getBaseUrl(HttpServletRequest request) {
+        try {
+            UrlPathHelper urlPathHelper = new UrlPathHelper();
+            URL url = new URL(request.getRequestURL().toString());
+            String host = url.getHost();
+            String userInfo = url.getUserInfo();
+            String scheme = url.getProtocol();
```
fxthomas: This currently doesn't handle the quite common case where a proxy is used to terminate HTTPS connections. This results at best in warnings such as this, or at worst in broken links :

screenshot from 2017-02-27 22-57-49

I quite like how Libresonic currently just grabs the currently-used protocol, this makes my proxy work without (much) additional configuration, but I recognize that changing `/dwr/` links to the full path is a pretty wonky way of doing things.

I can see two options here :

* Either add a configuration parameter for the complete base URL (incl. context path).
* Or, specifically document that we accept `X-Forwarded-Proto` (or equivalent) to determine the protocol if present.

I tend to prefer the former (since it would be an explicit step in the Libresonic configuration), @jooola mentioned he prefers the latter in #242. What do you think?

Either way this should go into the proxy docs.

EugeneKay: X-Forwarded-Proto is my preference for handling this case. Any configuration of SSL, etc should occur outside of Libresonic, not inside. Remember that it is possible to have multiple different endpoints proxying to the same backend instance(I have a use case for this).

fxthomas: That's a valid point. Let's go with this!

I still don't like that `X-Fowarded-Proto` isn't supported out of the box by proxies, but at least if it's well documented it'll be our own convention, and enough people seem to be using it that it's practically a de-facto standard.

libresonic-main/src/main/java/org/libresonic/player/service/NetworkService.java
```diff
@@ -53,53 +37,16 @@
...
-            return;
-        }
-
-        urlRedirectionStatus.setText("Idle");
-        if (urlRedirectionFuture != null) {
-            urlRedirectionFuture.cancel(true);
-        }
-        testUrlRedirection = test;
-        urlRedirectionFuture = executor.scheduleWithFixedDelay(urlRedirectionTask, 0L, URL_REDIRECTION_DELAY, TimeUnit.SECONDS);
+        portForwardingStatus.setText(NOT_SUPPORTED);
```
muff1nman: Good point, I thought I was used by the RESTController, but it isn't so is good to remove.

libresonic-main/src/main/java/org/libresonic/player/service/NetworkService.java
muff1nman: Yep, I'm following. It was silly of me to not test https redirect as well. Will update.

muff1nman:
@fxthomas I tested https with nginx and http with httpd. Would you be able to test haproxy with the latest changes? Take a look at the PROXY.md doc.

muff1nman:
Here is some documentation I found in reference to using `X-Forwarded-Host`:
https://httpd.apache.org/docs/2.4/mod/mod_proxy.html#x-headers
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host
https://en.wikipedia.org/wiki/List_of_HTTP_header_fields#Common_non-standard_request_fields

It doesn't sound like it is OOTB for haproxy, so we might need to add something like this:

    http-request set-header X-Forwarded-Host %[req.hdr(Host)]

as suggested [here](http://notes.iopush.net/configure-haproxy-in-reverse-proxy-with-http-authentification/)

carnifux:
Probably not the correct place to write this, but this commit spams the logs pointlessly and consistently. 

```[3/13/17 2:54:08 PM UTC]	DEBUG	NetworkService	Could not calculate proxy uri
[3/13/17 2:54:08 PM UTC]	DEBUG	NetworkService	Calculated base url to http://192.168.1.68:8080/subsonic/
[3/13/17 2:54:09 PM UTC]	DEBUG	NetworkService	Could not calculate proxy uri
[3/13/17 2:54:09 PM UTC]	DEBUG	NetworkService	Calculated base url to http://192.168.1.68:8080/subsonic/
[3/13/17 2:54:19 PM UTC]	DEBUG	NetworkService	Could not calculate proxy uri
[3/13/17 2:54:19 PM UTC]	DEBUG	NetworkService	Calculated base url to http://192.168.1.68:8080/subsonic/
[3/13/17 2:54:20 PM UTC]	DEBUG	NetworkService	Could not calculate proxy uri
[3/13/17 2:54:20 PM UTC]	DEBUG	NetworkService	Calculated base url to http://192.168.1.68:8080/subsonic/
[3/13/17 2:54:30 PM UTC]	DEBUG	NetworkService	Could not calculate proxy uri
[3/13/17 2:54:30 PM UTC]	DEBUG	NetworkService	Calculated base url to http://192.168.1.68:8080/subsonic/
[3/13/17 2:54:30 PM UTC]	DEBUG	NetworkService	Could not calculate proxy uri
[3/13/17 2:54:30 PM UTC]	DEBUG	NetworkService	Calculated base url to http://192.168.1.68:8080/subsonic/
[3/13/17 2:54:40 PM UTC]	DEBUG	NetworkService	Could not calculate proxy uri
[3/13/17 2:54:40 PM UTC]	DEBUG	NetworkService	Calculated base url to http://192.168.1.68:8080/subsonic/
[3/13/17 2:54:40 PM UTC]	DEBUG	NetworkService	Could not calculate proxy uri
[3/13/17 2:54:40 PM UTC]	DEBUG	NetworkService	Calculated base url to http://192.168.1.68:8080/subsonic/
```

muff1nman:
@carnifux this should be just a temporary pain point due to the following:

1. We don't have a trace level and in addition don't have a way to change log levels on a class basis currently due to the fact we have a custom logger. We need to change this at some point (i.e. #94)
2. DEBUG is only shown for beta or development releases. It will not be shown in the 6.2 release.

gilgolan:
Hi guys, 
I'm having some redirection issues when using traefik as reverse proxy engine.
I am redirecting https frontend to http backend (libresonic), in 6.2 version : 
- browsing to **https**://privateHost:privatePort/libresonic
- after entering user/passwd the browser performs **POST** with relevant content, response header will have the following location: **http**://privateHost:privatePort/libresonic/index.view
- now I already have the cookie so I just need to change http to https and the libresonic site loads completely on https except for some issues with settings, logout and getting started frame. 
- clicking logout for example will throw me back to **http**://privateHost:privatePort/libresonic/login?logout .

external players and anything else works fine, just login/logout and enter settings is broken (and requires manual fixing of address). do you have any idea why it is broken only on few actions?

-g

muff1nman:
It sounds like you're missing `X-Forwarded-Proto`. See https://airsonic.github.io/docs/proxy/prerequisites/.

PR #261, opened 2017-02-26
Updating Haproxy documentation

jooola:
I tested the old configuration but i wanted to simplify the configuration file !

### DIFF
--- a/documentation/PROXY.md
+++ b/documentation/PROXY.md
@@ -98,21 +98,22 @@ The following configuration works for Apache (without HTTPS):
 
 ### HAProxy
 
-The following configuration works for HAProxy (HTTPS with HTTP
-redirection):
+The following configuration works for HAProxy (HTTP and HTTPS):
 
 ```haproxy
-frontend https
 
+defaults
+    
+    # Use HTTP protocole
+    mode http
+    
+frontend https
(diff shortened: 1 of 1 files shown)


fxthomas:
The goal for this documentation is that we can provide a working configuration that people can safely copy-paste and expect to work with the minimum amount of changes, even if they're not familiar with HAProxy (which is my case, incidentally).

So, two things :

* I'm thinking we should at least mention `mode http` somewhere. It wasn't the default when I worked on #242, so people new to HAProxy can forget or accidentally delete it from the configuration. I'm fine with adding a `default` section though, at least it's explicit!
* The text above the configuration mentions "HTTP → HTTPS redirection". Is that still the case after removing the `ssl_fc` check? If not the text should be updated to say e.g. "HTTP and HTTPS".

jooola:
Yeah sure, but want to prevent having 2 same line in the configuration files, but you're right i'll add a default section !

We could keep the "forcing https" lines as an option ! This is best i think !

I'll correct those 3 lines !

fxthomas:
Awesome! As far as I am concerned this can be merged then :)

jooola:
Great, i'll do some others cosmetic !

PR #262, opened 2017-02-28
Update default themes

jooola:
I don't know much in Java, so i help where i can !
I changed some icons in default_themes (the more icons was a cellphone so i changed it to "+").

I removed some pictures/links in the more page. The android apps picture and the jamstash picture. They already have a link in the more page, and libresonic.org/pages/apps.jsp don't exist yet.


EugeneKay:
Why are you removing the More page icons? I don't see any reason for those to be removed at this time.

In the long term the More page should be rewritten and/or combined with the About page, but I don't think this is the right way of going about it.

jooola:
You mean the 2 screens for Android and jamstash ? For those i juste removed them beacause they where ugly and they already are links in the page ( for Android apps and jamstash)

I juste remplaced the smartphone icons from to top menu with a better one !

I can remove the commit about jamstash and Android apps if you want !

jooola:
So should i remove this PR ? or remove the commit that removes the 2 pictures on the more page ?

The interesting part of this PR is replacing the more icon with a cross icon

Issue #263, opened 2017-02-28
Favicons


jooola:
We need to change the favicons, so i suggest i propose some and we choose ?
I someone has an idea i take it !

![favicon_blue](https://cloud.githubusercontent.com/assets/19195485/23412601/52ad2aa0-fdd6-11e6-9470-24589fde982a.png)

![favicon_dark](https://cloud.githubusercontent.com/assets/19195485/23412301/6b3d041a-fdd5-11e6-990f-764ce90cba6a.png)

jooola:
Waiting to solv #192 before posting new ones but i guess i'll take the blue one

PR #270, opened 2017-03-04
Change logo and icons Fix #192 and #263

### LINKED ISSUE #192: New logo uses non-free fonts
Looks like the new logo is using [DIN Condensed](http://www.paratype.com/pstore/default.asp?fcode=PT_DNC&letter=D) which is a non-free font.

There may be some alternatives: [Alte Din](http://www.1001fonts.com/alte-din-1451-mittelschrift-font.html) or [Cabin](https://fonts.google.com/specimen/Cabin+Condensed?selection.family=Cabin|Cabin+Condensed)

I tried updating the logo with these, but my svg skills aren't the best - I couldn't quite achieve the same look: 

logo_dark svg
logo_light svg

the first is Cabin and the second in Alte Din.

jooola:
Fix #192 and #263


muff1nman:
This is missing some files I think:
- contrib/assets/logo-square_dark.svg
- contrib/assets/logo-square_light.svg
- contrib/assets/logo_dark.svg
- contrib/assets/logo_light.svg

muff1nman:
Nevermind that last comment, they were deleted.

jooola:
Yes, note that i build the SVG by converting the PNG files. So they are a bit more heavy...
I didn't know how to use Inkscape to reproduce the exact same pictures.

muff1nman:
Inkscape should be able to save the SVG's directly, it would be good to see if you can get these originals so we can resize images later.

jooola:
I meant that i could not reproduce the pictures using Inkscape, since i really don't know how to use this software !

I used Inkscape to import the PNG files and then save to SVG.
So if you see the SVG files they are heavier than the same pictures build from scratch ( using Inkscape ).

I may not understand your last comment but :  
* If you're looking for the originals i uploaded all the pictures in xcf ( gimp ) and png, with a really decent resolution.
* If you're talking about using the old SVG's with nonfree fonts i don't get why ?!

muff1nman:
Ideally I would have liked to see svgs either made from the original or made from scratch. They scale better.

Issue #276, opened 2017-03-06
Jukebox improvement


biconou:
I use Subsonic for streaming music on my Android phone but also to listen to music on my Hi-Fi stereo system. 
That's why I am very fond of the Jukebox player. 
Unfortunately, the legacy jukebox player is (IMOO) poorly developed. 
In past years, I developed for myself a better jukebox technology based on the [cmus](https://cmus.github.io/) audio player;  it works perfectly but has also downsides :

- cmus is only available on linux
- it requires a cmus process to be running
- it brings some hairy synchronization problems between Subsonic play queue

At the moment, I am working on a new improved jukebox player that aims to solve some of the legacy jukebox weaknesses. 
My new jukebox seems to work properly and I am on the way to submit a pull request for experimental purpose.
The new jukebox is based on the Java Sound API thus there is no dependencies on ffmpeg. It should work on both linux and windows and take every advantage of audio devices. 

For discussion, I propose the following list of all the legacy jukebox downsides that could be considered as a first attempt to build a roadmap around this development.

- no gapless play
- systematically resamples (through ffmpeg command line) thus it can not take advantage of Hi-Res medias
- depends on ffmpeg
- not able to scroll to a desired position
- can not deal with several audio devices (in case you have multiple audio cards plugged)
- difficult to drive from android devices

texneus:
Definately interested in seeing this succeed!  If you need a beta tester I'm in.  Only caveat is I probably can't do much with this for a few months yet (rebuilding my server currently, HTPC is next)

PR #277, opened 2017-03-06
Experimental improved jukebox [not ready to be merged]

biconou:
According to #276 this is a WIP experimental new jukebox based on Java Sound API. 

- does not use ffmpeg
- plays gapless
- plays Hi-Resolution audio files in Hi-Resolution mode if the audio device supports it
- supported formats : mp3, flac, ogg
- new UI that allows to position in a song

To use the new Jukebox, create a player and choose the "Java jukebox" technology instead of the "Jukebox" one.

Note that this jukebox player uses a java audio player I started to develop here : https://github.com/biconou/AudioPlayer


biconou:
@muff1nman @EugeneKay  This PR is now reasonably stable to be tested and (I think) can be added to a 6.3 roadmap

muff1nman:
I don't think its a good idea to add anything else into the /repo folder. We are trying to get rid of it, not add to it. All dependencies should go through maven central.

biconou:
@muff1nman  I agree. I'll find a way to publish my AudioPlayer librery to a central repo for the final release of that PR.

biconou:
@muff1nman 
Because Libresonic version has moved to 6.3 and I removed the AudioPlayer library from the local repo, I think that this PR is ready to be merged. 
The new jukebox player is stable enough even though a lot of improvements and optimisations still have to be done. But the core feature is there and ready to be tested.

libresonic-main/src/main/java/org/libresonic/player/ajax/PlayQueueService.java
```diff
@@ -612,18 +613,33 @@ public PlayQueueInfo sortByAlbum() throws Exception {
...
     }
 
+    private void updateJukebox(Player player, int offset) throws Exception {
+        if (player.getTechnology() == PlayerTechnology.JUKEBOX) {
+            jukeboxService.updateJukebox(player, offset);
+        } else if (player.getTechnology() == PlayerTechnology.JAVA_JUKEBOX) {
+            jukeboxJavaService.updateJukebox(player, offset);
+        }
+    }
+
```
muff1nman: This and other occurrences of this check seem to indicate a violation of dependency inversion. Can we not create an interface for interacting with either technology?

libresonic-main/src/main/webapp/WEB-INF/jsp/javaJukeboxPlayerControlBar.jspf
```diff
@@ -0,0 +1,35 @@
...
+                <div id="javaJukeboxVolumeSlider" style="width:80px;height:4px"></div>
+            </td>
+        </tr>
+    </table>
+</div>
+
+<script type="text/javascript" src="<c:url value='/script/playQueue/javaJukeboxPlayerControlBar.js'/>"></script>
+<script type="text/javascript">
+    initJavaJukeboxPlayerControlBar();
+</script>
```
muff1nman: jspf? Can we keep the names consistent and use jsp?

libresonic-main/src/main/java/org/libresonic/player/ajax/PlayQueueInfo.java
```diff
@@ -31,6 +31,7 @@
 public class PlayQueueInfo {
 
     private final List<Entry> entries;
+    private final int index;
```
muff1nman: Why do we need this new field?

muff1nman:
A rebase will be required so we dont end up with the history of those jar files.

biconou:
@muff1nman Thanks for this review. I'll treat each point.

libresonic-main/src/main/webapp/WEB-INF/jsp/javaJukeboxPlayerControlBar.jspf
biconou: .jspf is the standard extension for JSP Fragment

http://stackoverflow.com/questions/2081005/what-is-jspf-file-extension-how-to-compile-it

libresonic-main/src/main/java/org/libresonic/player/ajax/PlayQueueInfo.java
biconou: It is needed by the playQueue.jsp view related to the new Jukebox player to know what is the index of the song in queue currently played. It is used by the nowPlayingInfos service callback.

libresonic-main/src/main/java/org/libresonic/player/ajax/PlayQueueService.java
biconou: Agree that it's worth improving.

biconou: I didn't want to reorganize too much things here. I think we should but it's not the purpose of this PR.
I think that the updateJukebox method of the PlayQueueService was originally a quick and dirty workaround used to quickly add the jukebox feature in Subsonic. But I think that this jukebox feature was not really well designed and that's why it is weak.
The PlayQueueService should not have an updateJukebox method but instead we should use the play, pause, stop etc ... methods to keep PlayQueueService consistent across all players technologies.
To enforce this idea I also consider that the udateJukebox method is particularly messy as it is the unique entry point to all commands available on a jukebox player.
I've thought of reorganizing all these things but I think it's more worth doing when we change the playQueue UI replacing the PlayQueueService which is currently bound to DWR.

biconou:
@muff1nman I removed any tracks of local repo additions by filtering my git repo with 
`git filter-branch --tree-filter 'rm -rf repo/com/github' HEAD`

biconou:
This PR can be discarded as I created #446

PR #280, opened 2017-03-08
Avoid direct dependency on tomcat

### LINKED ISSUE #279: Libresonic 6.2 Beta1 breaks instances running with jetty.
Breaks instances that were previously running the war with jetty. Please find an attached log of the error.
[Libresonic.2017.03.06.log.txt](https://github.com/Libresonic/libresonic/files/826225/Libresonic.2017.03.06.log.txt)

The issue is the release does not include the a traditional war anymore like previous releases, instead the war now embeds tomcat.

Issue #142 mentions libresonic.war.original, which was not included in the release that would more than likely fix the issue.

To be honest with you though, I'm perfectly fine with how libresonic is now packaging the war, I believe it is an improvement. I just think there was some documentation missing. After reading issues and merges I was able to discover the architectural changes made to the war and adjust my java command accordingly.

For documentation purposes I'm including [the fix](https://github.com/hurricanehrndz/docker-containers/commit/5cead5fa9e81add61077feef83e48afb6451b2c4) I applied to my docker container in order to get libresonic up and running again.

You will notice in the merge, I no longer use jetty which I'm okay with. In fact, the only reason I used jetty before was because libresonic did not have it embedded like subsonic did. Now that libresonic is embedding tomcat, my need for using jetty no longer exists.

muff1nman:
Use reflection to invoke tomcat configs. This ensures that we dont have a direct dependency on tomcat and can run on any application server, not just tomcat. Fixes #279 

### DIFF
--- a/libresonic-main/src/main/java/org/libresonic/player/boot/Application.java
+++ b/libresonic-main/src/main/java/org/libresonic/player/boot/Application.java
@@ -1,10 +1,8 @@
 package org.libresonic.player.boot;
 
 import net.sf.ehcache.constructs.web.ShutdownListener;
-import org.apache.catalina.Container;
-import org.apache.catalina.Wrapper;
-import org.apache.catalina.webresources.StandardRoot;
 import org.directwebremoting.servlet.DwrServlet;
+import org.libresonic.player.Logger;
 import org.libresonic.player.filter.*;
 import org.libresonic.player.spring.LibresonicPropertySourceConfigurer;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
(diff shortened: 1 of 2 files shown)


muff1nman:
@fxthomas @biconou I'm curious what you guys think about this one. The other way we could go, is just to say we require tomcat and leave it at that.

muff1nman:
To expand a bit, I don't think it makes sense to do this just for Jetty, as that is why we have spring boot in the first place. But in theory, not just Jetty is affected, but any non-tomcat based app sever. So if for whatever reason you wanted to use WebLogic, that would potentially also be affected. I think it might be best to do this change to ensure we aren't too reliant on tomcat and that we only have a hard dependency on the servlet spec.

fxthomas:
I intended to experiment with Jetty at some point for performance, so I definitely welcome this! I have no issues with how it's done here, so have my 👍 !

Issue #114, opened 2016-07-19
Do not load preview of all folders in current folder


vestingz:
Currently, Libresonic seems to be generating a list of all available subfolders containing playable files in "Playing View", which are being displayed below the currently playing release. This causes Libresonic to substantially lag on bigger collections (checked for folders containing 1000+ subfolders). Is there any setting to disable this behaviour?

EugeneKay:
Can you provide an example URL of the frame or take a screenshot of the page that you are describing? I'm not certain exactly which view it is that you're experiencing issues with. I have not run into serious performance problems with my library(~150GB / 23,000 entries), but I also have fast hardware and disks.

vestingz:
![bildschirmfoto vom 2016-07-19 23-31-14](https://cloud.githubusercontent.com/assets/19933361/16967615/11079472-4e0a-11e6-90c8-8e0e3b46a8c6.png)
![bildschirmfoto vom 2016-07-19 23-34-27](https://cloud.githubusercontent.com/assets/19933361/16967614/11045d2a-4e0a-11e6-8c48-b37e00ec6ec4.png)
The URL would be /libresonic/index.view.
Libresonic/Tomcat is running on 2.8GHz/8GB Ram/SSD, with all the files stored on a 3-Disk-RAIDZ-Array. But indeed, accessing this view causes almost 100% CPU load on my laptop for a few seconds.

DaveWut:
If I'm not wrong, why this is lagging on your laptop and not on the server is because the list of folders and files to play are generated using javascript. This causes to make your browser load and render everything and not the server. I have a folder containing more than 2800 songs and I have the same problem. Plus, scrolling into this page lags a little. 

Excepts if the rendering methods are rewritten, this won't change for certain period of time. Maybe if you browse your library using another computer (and a more powerful one) it might be better.

vestingz:
Yes, totally makes sense considering my server barely showing any activity while i access these pages.
Other then rewriting the rendering methods, wouldn't it be possible to disable the folder listing/reduce it to a default amount of five (eg most similar) with the option to load more?

DaveWut:
This feature is partially implemented on Subsonic 6.0. When the folder contains too much sub folders, it will display a partial list, then offers the user to show the rest using a button (Show all albums): https://i.imgur.com/fWZY7ob.png. Maybe the same thing can be implemented and could be done too when a folder contains too much songs.

vestingz:
[...]

PR #292, opened 2017-03-15
Change to server-side pagination for Sibling Albums and Sub Dirs

muff1nman:
- Only paritial pagination, first X, then showAll
- Default is 40 for subDirs and sibling albums each
- showAll parameter causes page refresh with entirely new list
- Setting is a personal/user setting for each user saved in the database

TODO/Future improvements include:
- Finer grain pagination
- Pagination done at the database level
- Seamless refresh with JS
- Unify all pagination uses across the site

Address #114 

It is a pretty significant performance boost client side for large libraries in certain cases.


muff1nman:
Any feedback on this? Biggest question in my mind is if it makes sense to make this a user or system wide configuration.

Issue #290, opened 2017-03-14
Subsonic logger is boring


biconou:
In javadoc of org.libresonic.player.Logger one can read : 

>  Note: Third party logging libraries (such as log4j and Commons logging) are intentionally not
>  used. These libraries causes a lot of headache when deploying to some application servers
>  (for instance Jetty and JBoss).
> 

I do not agree. 

I propose to use slf4j as logging API without having to rewrite all log statements.

We could deprecate org.libresonic.player.Logger and redirect every of its methods to log using slf4j.

muff1nman:
The biggest problem I foresee with this is still logging to ${libresonic.home}/libresonic.log. It'll be a bit of a pain with slf4j. It might be possible to reimplement the same logic we use to decide where libresonic.home is using logbacks [conditional logic](https://logback.qos.ch/manual/configuration.html#conditional). But then we have to deal with the directory not being created at the time the logger starts running. Perhaps there is some programmatic configuration we can do...

Then again, maybe we just stop using libresonic.log all together and throw it all to stdout. Thats where all the other logs are going anyways. Maybe we just cleanup stdout a bit and put everything there?

fxthomas:
Is it possible to both store log entries in `libresonic.log` _and_ output them to the standard stream? The log file is cleared everytime Libresonic restarts, which makes it hard to keep traces in case I need them.

muff1nman:
Outputting to both places should be possible. The biggest hurdle is outputting to libresonic.log before we know where the home directory is and where it's placed at. I'll fiddle with it because what you describe would be ideal.

muff1nman:
The programmatic logback configuration might be nifty. However, it is a bit complex to setup with spring-boot, so this is probably going to have to wait until after 6.2

muff1nman:
Scratch that last comment, it turned out not complicated at all :)

biconou:
Thanks @muff1nman

PR #310, opened 2017-03-23
Consolidate Logging

muff1nman:
- Use sl4fj as a backend to org.libresonic.Logger
- Output the same logs to libresonic.log as the console
- Use spring-boot logging constructs
- Turn down logging to error for non-libresonic classes info for
  libresonic classes and liquibase (perhaps change this in the future,
  but might be helpful for folks migrating their databases).


muff1nman:
Addresses #290

PR #315, opened 2017-03-24
[DEV Only] Jetty Embedded Support

muff1nman:
Users, please ignore this. We are still going to ship with tomcat embedded for the near future, this just adds jetty for devs to evaluate any benefits. You can still deploy the distributed War onto Jetty if desired.

Spring boot supports both Jetty and Tomcat, however only one is supposed
to be used at a time. This is problematic for us, because we would like
to have both on the classpath so we can configure them (i.e.
org.libresonic.player.boot.TomcatApplication). To remedy this, we mark
both as provided, but have two profiles which then tell spring to
exclude one or the other from the lib-provided war. These exclude rules
are a bit fragile, but can be reproduced by analyzing mvn
dependency:tree output.

### DIFF
--- a/libresonic-main/pom.xml
+++ b/libresonic-main/pom.xml
@@ -45,15 +45,17 @@
         <dependency>
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-web</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.springframework.boot</groupId>
+                    <artifactId>spring-boot-starter-tomcat</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
         <dependency>
-            <groupId>org.springframework.boot</groupId>
-            <artifactId>spring-boot-starter-tomcat</artifactId>
-            <scope>provided</scope>
-        </dependency>
-        <dependency>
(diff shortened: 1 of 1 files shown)


biconou:
It's confusing. 
The following documentation (https://docs.spring.io/spring-boot/docs/current/reference/html/howto-embedded-servlet-containers.html#howto-use-jetty-instead-of-tomcat) suggest that if you want to use Jetty instead of Tomcat, you must exclude `spring-boot-starter-tomcat` from `org.springframework.boot` and then add `spring-boot-starter-jetty`.
In our configuration we exclude   `spring-boot-starter-tomcat` from `org.springframework.boot`  and then add both `spring-boot-starter-tomcat` and `spring-boot-starter-jetty`. 
Finally, jetty is ignored as tomcat is the prefered servlet container for Spring Boot.
I tried to test remove  `spring-boot-starter-tomcat` to force using jetty but it simply does not work.

muff1nman:
Have you tried using the profile switches? The idea was to make it so that one could toggle between tomcat/jetty with the maven profile setting.

biconou:
No, I did not. Sorry I should have read the feature description more carefully.

PR #325, opened 2017-03-29
Update HAProxy documentation

joshuaboniface:
This PR suggests two changes to the HAProxy documentation:

1) Replaces the `reqadd` statements with `http-request set-header` statements, including a dynamic `X-Forwarded-Host`. Rationale is that this is  cleaner (no manual adding of domain), and extensible to a multi-domain environment. Note that I have NOT tested the `:[port]` trick on this form of the command since I run on port 443 - testing here would be appreciated.

2) Moves the `redirect scheme` from the backend to the frontend. I disagree with the logic of the commit (002275468a6fa8e48b7b837fbb5eb4135af79a89) that moved this to the backend - in a multi-backend environment, putting it in the frontend is cleaner, and in this simple config it makes no functional difference.

### DIFF
--- a/documentation/PROXY.md
+++ b/documentation/PROXY.md
@@ -131,10 +131,12 @@ frontend https
     bind :80
     bind :443 ssl crt /etc/haproxy/certs/cert_key.pem
     
-    # Add X-Headers necessary for HTTPS
-    # Replace frontend.example.com with your SSL host and include :[port] if not running on port 443
-    reqadd X-Forwarded-Host:\ frontend.example.com
-    reqadd X-Forwarded-Proto:\ https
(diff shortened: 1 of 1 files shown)


jooola:
I agree with your 2. point !
I do have myself this configuration. I changed this because i didn't want to force SSL for every backend since  you don't always need all backend secured.

joshuaboniface:
That makes sense, this is one of those "it depends on the rest of your config" things - the beautiful flexibility of HAPRoxy! But I think for this simple config that (I'm confident) people will copy+paste, it makes more sense logically in the frontend, and plus it encourages people to only-SSL more things they might add in the future ;-)

Issue #326, opened 2017-03-29
Access to media is open to unauthenticated users   via  /stream URL


izderadicka:
Libresonic 6.2.beta2 – March 24, 2017

Access to URL `https://my.server/libresonic/stream?player=1&id=717`   is not authenticated and is open for everybody.  Thus anybody can easily scan and download my media collection.  
Just by listing  id from 1 to some maximum number (as id is assigned sequentially and for each successful response download media, anyone has access to my  full collection.

I think this is considerable security issue.

Current version of subsonic has solved it by adding auth parameter to  URL query string , which contains some kind of user authentication token.

muff1nman:
We may need to get creative with this so that sharing, Sonos and upnp isn't broken. Perhaps for those cases we use JWT as hinted at. It's not exactly an easy change, so might be a bit of a stretch to get this into 6.2. It might be worth delaying 6.2 just for this. @eugenekay thoughts?

izderadicka:
For UPnP/DLNA  ( and I guess for Sonos too) the intended use in on local LAN -  so this can be streamed over http locally with full access.  My concern is if I enable https access to libresonic ( and open port on my router). nothing should be accessible without proper authentication.  

Looking into dwr response for PlayQueueService.play.dwr there are actually two keys for stream - streamUrl and remoteStreamUrl -   my understanding would be that streamUrl is for local LAN only (and can be used in DLNA etc.) and remoteStreamUrl is for access over Internet - in this case I'd expect some authentication token. So there could be an utility to distinguish between local links and remote links? But current subsonic seams to user token for both, which may not be needed, as least from my perspective,

Also URL `/coverArt.view?id=742` has exactly same issue it's openly accessible to everybody.

muff1nman:
I really dislike the distinction between local and remote. I feel like it's too opinionated on the security model and fragile. I'd rather have everything be secure by default than having local be less secure than remote. Complexity is the biggest enemy of security.

muff1nman:
In develop now so am closing this as resolved, but please do test and verify.

PR #328, opened 2017-04-01
Lock down Unsecured Urls

muff1nman:
Addresses #326.

This is still in draft status (needs code formatting/cleanup/reorganization), but I figured I would open a PR in case anyone has any early feedback. This turned out to be a bit more complex than I was hoping, so would also be curious if anyone has any other ideas.

One of the main decisions made was to isolate share urls from each other. I.e. If someone is given one share url, should they be allowed to access all other share/stream/coverArt urls (if they knew the url endpoints). Perhaps trying to guard against this has increased the complexity here, but I think it is worth it.

TODO:
- [X] Test upnp
- [X] Test HLS
- [ ] Improve handling around auth failures of share urls
- [X] Code Cleanups/Formatting/Reorgs
- [X] Test Token expiration
- [X] Use Share URL expiration times
- [ ] Find a better way to handle JSP stateless auth (JWTAuthenticationProvider 48-50)
- [X] See if token can be shortened at all. (Not worth the extra effort/complexity).
- [X] Rebase


muff1nman:
I think this is fine to be merged now pending any reviews.  I'll leave this PR open for a couple days to give anyone who is interested the chance to do so. The remaining two open items can be addressed as needed later.

Issue #334, opened 2017-04-04
Link in login.insecure incompatible with CSRF token


ae-g-i-s:
After the installation, as long as the password for the admin account has not been modified, Libresonic displays the `login.insecure` message in `WEB-INF/jsp/login.jsp`
```
 57         <c:if test="${model.insecure}">
 58             <p class="warning"><fmt:message key="login.insecure"><fmt:param value="${model.brand}"/></fmt:message></p>
 59         </c:if>
```
In the English version and most others (see below), this text contains a link to `login.view?user=admin&password=admin`. However, **clicking on that link simply redirects back to the same login form**.

I was confused by this until I saw that the regular form has a CSRF token, which is not inserted into the link. Manually logging in as admin/admin worked.

I'm not sure what the best solution is; the easiest would probably be to change the text for these languages and remove the link. The usability impact is minimal and it removes the confusion I faced when it just redirected me back to the login page.
The second alternative is inserting the CSRF token into the link, but not all translations contain a link, so there's some potential for future breakage.
The third would be to disable CSRF token checking while the admin password has not been changed. While this would solve the issue, it's still kinda ugly to enable random crawlers to accidentally log in (but that would already be possible if this bug didn't exist).

This issue affects the following languages (derived by grepping the `login.insecure` key from the i18n resource bundles and additionally filtering on `href`); others omit the link and only describe the steps to login.
```
bg
ca
cs
da
de
el
en
et
fi
is
ja_JP
ko
nl
pl
pt
sl
zh_CN
zh_TW
```

muff1nman:
Lets just remove the link. Its not worth the extra hassle with the csrf token. Is this something you could submit a PR for?

ae-g-i-s:
Well, I can remove it from English, German and a few other translations and do a PR for that, but not sure about languages that aren't related to these (especially Slavic and Chinese).

muff1nman:
For any you're not sure about, leave them as is.

muff1nman:
> The following translations still contain the broken link: bg cs el ja_JP ko pl sl zh_CN zh_TW

We can leave this open as a placeholder

muff1nman:
I think we might just have to accept the fact that not all translations will be up to date.

PR #335, opened 2017-04-05
Remove broken login link from login.insecure message.

ae-g-i-s:
This fixes issue #334 for the following locales: ca da de en et fi is nl pt.
I have altered the wording of the second sentence in German too, as it was somewhat clumsy (switched from "friendly" to "formal" and included a capitalisation error).

The following translations still contain the broken link: bg cs el ja_JP ko pl sl zh_CN zh_TW


muff1nman:
Although this will say "closed", I actually cherry-picked this into the develop branch. See c6572ccc86c70323cfbc1191f021ce11702e8a9f. We should really change the default branch at some point.. At any rate, thanks a bunch!

PR #340, opened 2017-04-08
Add Moment.js library

biconou:
This PR adds Moment JS library (https://momentjs.com/) for date and time manipulations.


muff1nman:
What is this related to? Why split this out from the PR that needs this?

biconou:
This is related to a work I am doing on jukebox UI. It's used to calculate song duration.
I propose it here as a general purpose library to manipulate dates. So I created a dedicated PR for that. 
It can be considered as part of some guideline (should be documented) for javascript development.

muff1nman:
Makes sense and looks handy. But it doesn't make sense to merge this in by itself but as part of the jukebox PR.

biconou:
Ok. I'll discard this PR.

PR #358, opened 2017-04-22
Added a property to ignore file timestamps when reindexing

muff1nman:
#### Background
I accidentally deleted most of my music directory. The database was still intact. I recovered the music directory by rolling back to a previous ZFS snapshot and performed a reindex. However, libresonic did not mark the deleted files as present. Turns out the file timestamp was unchanged through the ZFS restore, and so libresonic still thought the last indexing effort was still "good".

#### Changes
This adds the option to ignore file timestamps when scanning files. This can be helpful in the case of a restore as described above. There might be a better way to do this, as this was really a quick effort on my part to fix my own libresonic.

#### TODO
This does not add a UI, just a single property that can be turned on by editing the lilbresonic.properties file.

### DIFF
--- a/libresonic-main/src/main/java/org/libresonic/player/dao/AlbumDao.java
+++ b/libresonic-main/src/main/java/org/libresonic/player/dao/AlbumDao.java
@@ -314,12 +314,12 @@ public class AlbumDao extends AbstractDao {
     }
 
     public void markNonPresent(Date lastScanned) {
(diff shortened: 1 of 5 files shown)


fxthomas:
Rather than a property, why not make it a GET parameter that can be changed at runtime in the `musicFolderSettings.view?scanNow` request?

muff1nman:
@fxthomas That's not a bad idea.

PR #359, opened 2017-04-22
Improve Playlist Handling

muff1nman:
- Use external library chameleon (lizzy)
- Adds the ability to specify playlist export format
- Fixes some deficiencies with playlist handling

~~Not yet ready because chameleon is not yet in the maven central.~~ Also doesn't yet include any ui.


jooola:
This will be for 6.3 or 6.2.1 ?

PR #372, opened 2017-05-01
Proxy fixes

NHellFire:
This fixes 2 issues with the current proxy handling:
1. If libresonic is behind multiple reverse proxies, the proxy headers would be ignored as other proxies will append their own headers.

2. Use the full URL when redirecting from nowPlaying.view.
Otherwise it could redirect from HTTPS to HTTP and then be blocked by the browser.

### DIFF
--- a/libresonic-main/src/main/java/org/libresonic/player/service/NetworkService.java
+++ b/libresonic-main/src/main/java/org/libresonic/player/service/NetworkService.java
@@ -60,8 +60,21 @@ public class NetworkService {
 
     private static URI calculateProxyUri(HttpServletRequest request) throws URISyntaxException {
         String xForardedHost = request.getHeader(X_FORWARDED_HOST);
+        // If the request has been through multiple reverse proxies,
+        // We need to return the original Host that the client used
+        if (xForardedHost != null) {
+            xForardedHost = xForardedHost.split(",")[0];
+        }
+
         if(!isValidXForwardedHost(xForardedHost)) {
             xForardedHost = request.getHeader(X_FORWARDED_SERVER);
+
+            // If the request has been through multiple reverse proxies,
+            // We need to return the original Host that the client used
+            if (xForardedHost != null) {
+                xForardedHost = xForardedHost.split(",")[0];
+            }
+
             if(!isValidXForwardedHost(xForardedHost)) {
                 throw new RuntimeException("Cannot calculate proxy uri without HTTP header " + X_FORWARDED_HOST);
             }


muff1nman:
The first change for multiple proxies makes sense, but I don't understand the need for the second change. The redirect works fine for me behind a single proxy. In addition, we have other places where there are relative redirects. I.e. all the settings pages. Would those need to be updated as well?

NHellFire:
For me, nowPlaying.view was redirecting to http://subsonic.domain.com/home.view, which then gets blocked by my browser because the rest of the page is HTTPS.

My proxy headers are correct (`NetworkService.getBaseUrl()` returns the correct URL, as well as netcat shows what I expect), but it seems all relative redirects are being transformed to HTTP.

Settings also has the same problem:
`$ curl -s -D- https://subsonic.domain.com/settings.view -b JSESSIONID=.... | grep Location:
Location: http://subsonic.domain.com/musicFolderSettings`
It's probably a side effect of my network setup (single IPv4, httpd on multiple servers).

muff1nman:
We are currently depending on the proxy (or in your case proxies) to rewrite the redirects appropriately. For apache, the `ProxyPassReverse` [claims](https://httpd.apache.org/docs/2.4/mod/mod_proxy.html#proxypassreverse) that it does the rewrite of redirects. Does a single httpd work?

NHellFire:
As far as I can tell, Apache won't rewrite redirects if they contain a hostname.
Spring adds the protocol and hostname itself:
`$ curl -b "JSESSIONID=...." -s -D- http://127.0.0.1:4040/settings.view | grep Location:
Location: http://127.0.0.1:4040/musicFolderSettings`

muff1nman:
Just tested with httpd and seems like it returned the relative path which was correctly redirected by the browser.
```
$ curl  -v 'http://localhost:81/libresonic/settings.view?'   -H 'Cookie: player-61646d696e=1; JSESSIONID=E3435E82DFE49AD43D4794F734B8F390; player-61646d696e=1' 
* About to connect() to localhost port 81 (#0)
*   Trying ::1...
* Connected to localhost (::1) port 81 (#0)
> GET /libresonic/settings.view? HTTP/1.1
> User-Agent: curl/7.29.0
> Host: localhost:81
> Accept: */*
> Cookie: player-61646d696e=1; JSESSIONID=REDACTED; player-61646d696e=1
> 
< HTTP/1.1 302 Found
< Date: Thu, 04 May 2017 21:57:13 GMT
< Server: Apache-Coyote/1.1
< X-Content-Type-Options: nosniff
< X-XSS-Protection: 1; mode=block
< Cache-Control: no-cache, no-store, max-age=0, must-revalidate
< Pragma: no-cache
< Expires: 0
< X-Frame-Options: SAMEORIGIN
< Location: musicFolderSettings
< Content-Language: en
< Content-Length: 0
< 
* Connection #0 to host localhost left intact
```

NHellFire:
Seems certain proxy settings confuse it (such as passing the Host header to libresonic). I've dropped that commit now.

PR #400, opened 2017-05-12
New assets

### LINKED ISSUE #399: Rebuild assets (logo and icons)
First creation of the logos and icons svg files were imported from xcf (gimp) files.
So this was converting  non vector graphics into vector graphics. Resulting in bad scalling svg files.

This time I started with creating the SVG and i'll convert it to PNG.

Here is a sample : ( hope this one don't differ the old one ! )
 
libresonic

#263

jooola:
This PR replace old assets with better ones.
Fix #399


muff1nman:
Looks like some of the sizes changed. Is there any reason for this?

muff1nman:
But overall, love having the svg's back and seems like some of the icons are now a bit crisper.

jooola:
I'm glad you like it !

The size changed a bit because i wanted to center the logo.
If you need specific size just export the svg in a size you need. The ratio may have changed hope this wont break anything. I didn't care too much about the ratio.

About the size of the logo in the left panel i wanted to use the free space that was left ( the width is the exact size of the picture ! )

muff1nman:
As dicsussed on IRC, here is an example of size changes:

original_1

vs

new_1

Where the original seems like it was left justified, the new one with the extra padding is somewhere confused between left justified and centered.

jooola:
This should good ! Tell me if this padding is right !

jooola:
#421 Fix this !

Issue #45, opened 2016-05-05
Move to HTML5 Player


EugeneKay:
Moving to a HTML5 player is desired. Patches welcome.

jolny:
Could http://jplayer.org/ be an alternative? It's under a MIT license.

csobsidian:
Native HTML5 audio? or are we talking video?

apastuszak:
I believe Ampache uses jplayer, but I could be wrong.

jolny:
@csobsidian it appears it supports both :)

csobsidian:
a list of html5 video/audio players and their FOSS licenses:

http://html5video.org/wiki/HTML5_Video_Player_Comparison

csobsidian:
It's more work but we could write our own interface and use the [Web Audio API](https://www.w3.org/TR/webaudio/), which is supported by [all current major browsers](https://en.wikipedia.org/wiki/HTML5_Audio#Web_Audio_API_and_MediaStream_Processing_API). This option would allow us to do bumpless playback by pre-buffering the next audio file, software EQ in browser, and even audio aware themes.

RomanHargrave:
@csobsidian's idea is my favorite

RomanHargrave:
To follow up, I just switched subsonic (the kang libresonic came from) to libresonic, and found to my (disgust) that this using a flash player. Good lord.

apastuszak:
The kang has always used as flash player, as has Subsonic prior to version
6.

RomanHargrave:
@apastuszak I was able to use it in a browser without flash support. It appeared to be using JWPlayer, which will use HTML5 audio if the browser supports it.

apastuszak:
You were probably running 6.0beta1 of kang.  I believe the kang version had
to roll back to 5.3 for the Libresonic release, since that was the last
version that was GPL licensed.

RomanHargrave:
On Thu, 12 May 2016 11:50:28 -0700
apastuszak <notifications@github.com> wrote:

> You were probably running 6.0beta1 of kang.  I believe the kang version had
> to roll back to 5.3 for the Libresonic release, since that was the last
> version that was GPL licensed.
> 
> On Thu, May 12, 2016 at 2:32 PM, Roman Hargrave <notifications@github.com>
> wrote:
> 
> > @apastuszak <https://github.com/apastuszak> I was able to use it in a
> > browser without flash support. It appeared to be using JWPlayer, which will
[...]

Issue #390, opened 2017-05-08
Chromecast is not working


Tapchicoma:
After updating to 6.2 I've checked if chromecast now working, but it is not.

I can start casting session, but streaming itself fails.

From chrome dev tools I see that chrome session is successfully started, but failed to load stream url:
`code: "session_error", description: "LOAD_FAILED"`

I've checked stream url and it is valid (at least browser opens it and plays song).

My suspicion is that this stream url doesn't allow other then current browser to access it and casting device so can't open it.

Tapchicoma:
After reverting commit f7bd43136b6275 casting is working again.

muff1nman:
Sounds like we are sending the plain stream url to the chromecast. We need to send one with the jwt token. I have a chromecast, but the earliest I can get to fixing it is this weekend.

muff1nman:
It should now work with audio. Video may still be broken and need some work.

Tapchicoma:
Checked, now it works perfectly. Would be nice to have something like 6.2.1 release with this bug fix included.

PR #425, opened 2017-05-20
Replaced jwplayer with HTML5 player using MediaElement.js

Eusmilis:
As discussed in #45, I moved to HTML5 player for audio and video.
I also changed the transcoding command to stream mp4 and not flv.


jooola:
Language string are provided by MediaElement right ? So it's a bad idea to move the strings to the l18n files ?
```
libresonic-main/src/main/resources/org/libresonic/player/i18n/
```

muff1nman:
Wow, thanks a bunch for taking a stab at this. I will review more in depth asap, but just one initial thought:

On the liquibase migrations, we need to be extra cautious in modifying the existing migrations. Liquibase is designed such that it will throw an error if migrations are previously modified as it no longer knows how to handle existing database files. Really it forces you into two options:
1. Modify the previous sql migrations, and add another valid checksum for the old migration. In this case, existing databases will ignore any changes to the sql migration files
2. Create new sql migrations that are designed to be applied to existing databases.

Option 1 we want to avoid as we want everyones transcodings to be updated. Option 2 is really the best way to go about it so everyone gets the proper updates and I think we can go that route here. For example, we might add a migration that looks like:
```
        <update tableName="transcoding">
            <column name="name" value="avi > mp4"/>
            <column name="source_format" value="avi" />
            <column name="target_format" value="mp4" />
            <column name="step1" value="ffmpeg -ss %o -i %s -async 1 -b %bk -s %wx%h -ar 44100 -ac 2 -v 0 -f mp4 -vcodec libx264 -preset superfast -threads 0 -movflags frag_keyframe+empty_moov -" />
            <where>name = 'avi > flv'</where>
         </insert>
```

Let me know if you need any help with the liquibase, I'd be glad to send a PR to your branch if you'd like to avoid the liquibase.

Again, this is just the first thing that jumped out at me, otherwise I'll take some time to test and review. Thanks again!

muff1nman:
@jooola I would assume that the mediaelement script files were copy pasted from a mediaelement download with no modifications. @Eusmilis correct? We definitely want to avoid doing any modifications to mediaelement to allow future upgradeability.

Eusmilis:
Thanks.
I never worked with liquibase before so I would appreciate some help.
You are right concerning the script files. I thought it best to leave everything related to MediaElement together and treat it like an external library.

jooola:
Ok great thanks for this !

Were could i get a WAR package with this PR included ? @muff1nman ?

muff1nman:
Okay I'll send a PR to your branch with what I'm thinking.

biconou:
Why is there so many js files in that PR ? It would be preferable to use a compact distribution of mediaelements in one unique js file if it exists.

muff1nman:
@biconou If you mean that we should use a minified version, I think that the debuggability is more valuable than a more compact version.

biconou:
@muff1nman debuggability is usefull in development conditions but a minified version would be better for production.

Eusmilis:
@biconou the files linked in the JSP are the minified version. I just left the development version to be able to debug if necessary.

biconou:
@Eusmilis ok

muff1nman:
Had just a bit of time to review this today and here are my initial thoughts:

### Transcoding
Basically, this is currently working because the flv transcode was replaced with the mp4 transcode and the mp4 transcode just happens to be the last transcoder that is selected rather than directly selecting a transcoding we want. So at a minimum, we need to add the mp4 transcodings (leaving the flv transcodings be) and set the desired format to get the right transcoding in the video stream url like this:
```
        String streamUrl = url + "/stream?id=" + file.getId() + "&player=" + playerId + "&format=mp4";
```
However, I think we could do better. For one, mediaelement supports many formats. Flash being one of them. In fact, we could potentially even use our existing flv transcoder by fixing the content type to match as specified [here](https://github.com/mediaelement/mediaelement/blob/master/docs/usage.md#use-renderers) (I think we currently use `video/x-flv`) without adding the mp4 one. I still think adding the mp4 one is a good idea, but really this shows how we're lacking a good way to negotiate on the transcoding to use with all the different formats mediaelement supports.

Soooo.. really what I would like to see done is deprecate the `format` parameter on the stream controller and use the actual HTTP `Accept` header that was designed for this use and also has better semantics for priority and multiple acceptable content-types. I.e. we could send:
```
Accept: video/mp4;video/flv
```
And the behavior would be that we use the first available transcoder on the server side. One thing to verify in the future is that if the file type is in the Accept list, we prioritize sending the file without transcoding provided no downsampling is desired.

### Chromecast / JSP / Javascript / Controls
Although these are kinda unrelated, its really one intertwined mess in the libresonic code base. To be explicit we have the current problems:
- Large amounts of JS from the `CastPlayer` intertwined with jsp code
  - also impacts ones ability to debug in browser
- `CastPlayer` seemingly duplicated across the playQueue.jsp and the videoPlayer.jsp files

The above are current problems that were there before this PR, but with this PR I think we also are missing some vital controls on the video player. The biggest being the inability to fullsize the video.

The good news is that I think this all can mostly be killed with one stone. Namely, we could migrate to the mediaelement chromecast support ([a plugin](https://github.com/mediaelement/mediaelement-plugins/blob/master/docs/chromecast.md)) and then hopefully simplify our current `CastPlayer` implementation and synchronize the implementation across both `playQueue.jsp` and `videoPlayer.jsp`. We would also enable the builtin mediaelement controls for the video player.

### Misc
Other thoughts are:
- A set of steps to reproduce the mediaelement sources in the libresonic source tree (i.e. in case of upgrade)
- We need to integrate the mediaelement styling into our themes

Now I understand this is a ton of work and I don't think it is realistic to expect this all to be done by @Eusmilis. Therefore I propose the following milestones to getting this to a stable release:

#### PR Merged
- [ ] Liquibase changes reverted
- [ ] New Liquibase migration for any new transcodings we want/need to use with mediaelement to at least get it working OR see if `video/flv` works
- [ ] steps for reproducing the mediaelement sources (i.e. in case of upgrade)


#### Develop Merged to Stable
- [ ] Remove dup code between audio and video jsp pages
- [ ] Move javascript embedded in videoPlayer/playQueue.jsp to standalone js
- [ ] Chrome cast support (use mediaelement plugin)
- [ ] Full screen control OR use ME controls for video (although I'd like to see this one done as part of the PR, it might be dependent on some of these other tasks)
- [ ] Stylizing of ME to match themes

#### Future work
- [ ] Support no transcode in the case none is needed.
- [ ] Change format type to use accept header (allows proper negotiation with server without a statically set format)
- [ ] Research into the most optimal video formats to use and prioritize them.

Although it would be nice to have @Eusmilis help where he/she can here, I would be willing to help with these tasks. My time is a bit limited over the next week or two, but there's no need to rush and I'd like to see this done the right way so we can live with this for a while.

muff1nman:
After writing that, I would actually be curious on @Eusmilis thoughts on the feasbility of the above road map. In particular, you're probably one of the select few familiar with the `CastPlayer` so I'm curious on if you think the audio and video implementations can be reconciled.

Eusmilis:
I think most of it can be done, but I'll have to take a closer look at the `CastPlayer` and the chromecast plugin for MediaElement to see how we could make it work.
For now, I didn't use ME for the video player since I wasn't sure how it would interact with `CastPlayer`.
I'll tell you more after I look at it.

I'm not sure what you mean by "steps for reproducing the mediaelement sources". Do you mean we should make MediaElement a git submodule ?

jooola:
I can help with the styling and all the visual content. But this isn't for right now !
I'll be around tell me when I can help !

muff1nman:
Maybe I'm being naive, but it wasn't readily apparent to me where you got the mediaelement sources from. Did you pull it from GitHub? Did you build it with Grunt?

Eusmilis:
I downloaded the release from github and copied the files.

muff1nman:
Okay I see it now in the build directory, and we can scratch that todo.

libresonic-main/src/main/webapp/WEB-INF/jsp/playQueue.jsp
```diff
@@ -680,8 +671,8 @@
             </c:if>
             <c:if test="${model.player.web}">
                 <td>
-                    <div id="flashPlayer" style="width:340px; height:24px;padding-right:10px">
-                        <div id="jwplayer"><a href="http://www.adobe.com/go/getflashplayer" target="_blank"><fmt:message key="playlist.getflash"/></a></div>
+                    <div id="player" style="width:340px; height:40px;padding-right:10px">
+                        <audio id="audioPlayer" class="mejs__player" data-mejsoptions='{"alwaysShowControls": "true"}' width="340px" height"40px"/>
```
jooola: We are running in an issue here: <audio> element only exist in HTML5, but if you load the frame in HTML4, the audio element won't be recognized !
See some links here : 
https://www.w3schools.com/tags/tag_audio.asp
https://www.w3schools.com/html/html5_intro.asp

I think that the following issue needs to come with this PR : #427

muff1nman: What browser are you using where it causes a problem?

jooola: I don't have any issue i didn't tested it.
What is say is that we can't add html5 element in html4.01 specially when the element is supposed to be new in html5.
I may be working but this isn't clean

image

image

image

muff1nman: As long as it works on the browsers we can address #427 separately.

muff1nman:
@Eusmilis what do you think about my changes in https://github.com/Eusmilis/libresonic/pull/1?

Also, if we head down the route of using the builtin mediaelement.js support for chromecast without modifications it might be simpler to just use the remote stream url everywhere. Even on audio. Although I was initially hoping to not use remote stream url when we have a valid cookie, but I dont think there is any huge downside to changing that.

muff1nman:
Also, I think we can use the builtin mediaelement js controls, but then add custom download and bit rate controls like https://stackoverflow.com/questions/28470695/how-to-add-custom-button-inside-html-player-using-mediaelement-js-script

Eusmilis:
@muff1nman seems good to me. I'm going to test it and merge it.

I did some tests with the chromecast plugin for mediaelement.js. Apparently, it can only cast video, there is no cast button on the audio player.

I'm trying to get it to work for the video, but I couldn't so far. It works well when the source is a file, but I get an error from the cast api with the stream url. I have the same problem with `CastPlayer`.

I had to apply your fix for #390 for the audio to work on the chromecast, so I think it has something to do with the jwt token missing from the video stream url.

muff1nman:
To simplify testing things, it might be good just to turn off security. We can always hash out those details later. I.e. I think you could just modify these lines here:
```
                    .antMatchers("/**")
                    .hasRole("USER")
```
Perhaps remove the hasRole line. 

Good to know about it not supporting media. That seems like it could be a sticking point for us really unifying things. The [plugin](https://github.com/mediaelement/mediaelement-plugins/blob/master/src/chromecast/chromecast.js) looks pretty well put together, perhaps we could see if its a simple change and/or see if the mediaelement js guys can help us.

muff1nman:
@Eusmilis looks good. Only other things I can think of is that we need to also update the external player and then remove the jwplayer sources. I added these to the list I created above, but I don't think they need to block us from merging this into develop. Do you think this is good to get merged?

Eusmilis:
I think we can merge into develop.

I will continue to work on the video and chromecast player, see if we could get rid of `CastPlayer` and use only MediaElement and its plugins.

For now, I managed to cast video using the chromecast plugin by disabling the security, but it only played what was buffered. If we want to use it, I think we will have to change the plugin to be able to set the duration of the video.

muff1nman:
It only playing what was buffered seems like a bug of a sort.

muff1nman:
With this pr merged, I think it makes sense to create a separate issue on things still todo.

PR #429, opened 2017-05-21
Change more icon ( phone icon > + icon )

jooola:
The more icon in the default theme is a phone...
I don't think this matches really the purpose of the More page !

This PR only change the default themes more icons with new ones.


jooola:
image

image

PR #431, opened 2017-05-21
New login page

### LINKED ISSUE #416: New login layout
This issue is a discuss about some changes on the login page.
PR is almost ready but maybe some things needs to be discussed !

Original : 
image

Here is a screenshot with the changes

image

jooola:
This PR is a rework of login page to fix #416 
See screens in the issue ticket.

### DIFF
--- a/libresonic-main/src/main/webapp/WEB-INF/jsp/login.jsp
+++ b/libresonic-main/src/main/webapp/WEB-INF/jsp/login.jsp
@@ -10,48 +10,50 @@
     </script>
 
 </head>
-<body class="mainframe bgcolor1" onload="document.getElementById('j_username').focus()">
+<body style="min-width:550px" class="mainframe bgcolor1" onload="document.getElementById('j_username').focus()">
 
 <form action="<c:url value="/login"/>" method="POST">
 
     <sec:csrfInput />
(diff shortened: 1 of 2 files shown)


libresonic-main/src/main/webapp/WEB-INF/jsp/login.jsp
```diff
@@ -10,48 +10,50 @@
     </script>
 
 </head>
-<body class="mainframe bgcolor1" onload="document.getElementById('j_username').focus()">
+<body style="min-width:550px" class="mainframe bgcolor1" onload="document.getElementById('j_username').focus()">
```
muff1nman: Wouldn't it be a good idea to push some of the css into the css files instead of embedding them in the html? Seems like a valid use case to be able to customize the login page with a theme css file.

muff1nman:
Aside from the one concern on organization on the css, everything looks good here. I think its a major improvement 💯

libresonic-main/src/main/webapp/WEB-INF/jsp/login.jsp
jooola: Ok yeah, i didn't wanted to differ too much from the original code, but i can do this yes !

jooola:
image

jooola:
I need a second review please

PR #438, opened 2017-05-26
Remove Ubuntu fonts and store Roboto locally

### LINKED ISSUE #398: Find new fonts and provide them locally
I think we should provide fonts locally instead of loading them from the google api

jooola:
Ubuntu fonts are only used in the `<h1>` element. I suggest to remove this font to speed up the loading time !

fix #398


README.md
```diff
@@ -37,6 +37,7 @@ Libresonic is free software and licensed under the [GNU General Public License v
 
 The [Subsonic source code](https://github.com/Libresonic/subsonic-svn) was released under the GPLv3 through version 6.0-beta1. Beginning with 6.0-beta2, source is no longer provided. Binaries of Subsonic are only available under a commercial license. There is a [Subsonic Premium](http://www.subsonic.org/pages/premium.jsp) service which adds functionality not available in Libresonic. Subsonic also offers RPM, Deb, Exe, and other pre-built packages that Libresonic [currently does not](https://github.com/Libresonic/libresonic/issues/65).
 
+Libresonic software includes Roboto Fonts by Google, distributed under GPLv3. Available as Apache 2.0 from [https://fonts.google.com/specimen/Roboto](https://fonts.google.com/specimen/Roboto).
```
muff1nman: Roboto Fonts is distributed under Apache 2.0 is what you meant to say?

libresonic-main/src/main/webapp/style/fonts.css
```diff
@@ -0,0 +1,79 @@
...
+
+/* roboto-700italic - greek_vietnamese_latin_cyrillic */
+@font-face {
+  font-family: 'Roboto';
+  font-style: italic;
+  font-weight: 700;
+  src: local('Roboto Bold Italic'), local('Roboto-BoldItalic'),
+       url('../fonts/roboto-v16-greek_vietnamese_latin_cyrillic-700italic.woff2') format('woff2'), /* Chrome 26+, Opera 23+, Firefox 39+ */
+       url('../fonts/roboto-v16-greek_vietnamese_latin_cyrillic-700italic.woff') format('woff'); /* Chrome 6+, Firefox 3.6+, IE 9+, Safari 5.1+ */
+}
```
muff1nman: Why is this file not identical to the file at `https://fonts.googleapis.com/css?family=Roboto:400,300,400italic,500,300italic,500italic,700,700italic,100,100italic`?

README.md
jooola: Yeah, exactly !
Roboto is Apache2.0 and we are using them locally.
Should I point the link to the Apache2.0 license instead of the font page ?
Sorry i don't know how i can write it in a better way.

libresonic-main/src/main/webapp/style/fonts.css
jooola:
```
<link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Ubuntu&subset=latin,cyrillic-ext,greek-ext,greek,latin-ext,cyrillic" type="text/css"/>
<link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto:400,300,400italic,500,300italic,500italic,700,700italic,100,100italic" type="text/css"/>
```

jooola: Sorry for the previous message (deleted)...

jooola: Sorry I didn't understand what you meant.

It is basically pretty much the same.
Only thing missing is the Ultra Thin fonts ( font-weight: 100 )
Since we aren't using such a thin font in Libresonic I didn't added it.

I used (like @fxthomas ) the [fonts download helper](https://google-webfonts-helper.herokuapp.com/fonts), and I think it is normal that the code change more than just a bit since we are providing them locally.
But it works the same!

* 2 fonts src for every fonts => Better compatibity
* fonts are sorted in a different way
* unicode range is used to specify which font use when a specific character is used [see here](https://24ways.org/2011/creating-custom-font-stacks-with-unicode-range) but since we write everything in Roboto we don't need it.

README.md
muff1nman: Okay I understand now.

libresonic-main/src/main/webapp/style/fonts.css
muff1nman: The files are drastically different, lets download it exactly. No modifications.

jooola: I don't understand what is wrong. The end point is the same.
I can add the "thin" fonts, remove the second src for the woff files.

But I don't see what is missing or useless.
The old files have a `@fontface` for every fonts language ( latin, cyrillic, ...) in the new one they are condensed in 1 file (per font-weight/font-style).

Please tell me precisely what's wrong!

muff1nman: I would like for this command to not show any difference:
```
diff -y  <(curl 'https://fonts.googleapis.com/css?family=Roboto:400,300,400italic,500,300italic,500italic,700,700italic,100,100italic') <(curl 'https://raw.githubusercontent.com/jooola/libresonic/7d80175d9497a16af5a360dbdccd759d74e97e7b/libresonic-main/src/main/webapp/style/fonts.css')
```

PR #446, opened 2017-05-30
New JukeBox

biconou:
This PR is for the new JukeBox (and replaces #277).
The new JukeBox works both from Android app or web UI.
This PR can be reviewed, tested and merged.

We could also wait for improvements I'd like to do. I'd like to use the Rest API to control the JukeBox from the web UI but it needs some work on Rest authentication before.

**Mini documentation**

To create a new JukeBox player choose the _Java Jukebox_ technology in the player settings page and select an available audio device (sound card) in the _Audio device_ combo.

sc1

The Java Jukebox UI player controller now has a slider to scroll while playing.

sc2

This control bar could be improved but I think it's worth change things in the playqueue UI before.

Enjoy.


libresonic-main/src/main/java/org/libresonic/player/service/JukeboxJavaService.java
```diff
@@ -0,0 +1,299 @@
...
+            throw new RuntimeException("The player "+libresonicPlayer.getName()+" is not a java jukebox player");
+        }
+        com.github.biconou.AudioPlayer.api.Player audioPlayer = retrieveAudioPlayerForLibresonicPlayer(libresonicPlayer);
+        if (audioPlayer != null) {
+            return audioPlayer.getGain();
+        }
+        return 0.5f;
+    }
+
+    public synchronized void setGain(final Player libresonicPlayer, final float gain) {
```
muff1nman: Why do we need synchronized here?

libresonic-main/src/main/java/org/libresonic/player/service/JukeboxJavaService.java
```diff
@@ -0,0 +1,299 @@
...
+     * Finds the corresponding active audio player for a given libresonic player.
+     * The JukeboxJavaService references all active audio players in a map indexed by libresonic player id.
+     *
+     * @param libresonicPlayer a given libresonic player.
+     * @return the corresponding active audio player of null if none exists.
+     */
+    private com.github.biconou.AudioPlayer.api.Player retrieveAudioPlayerForLibresonicPlayer(Player libresonicPlayer) {
+        com.github.biconou.AudioPlayer.api.Player foundPlayer = activeAudioPlayers.get(libresonicPlayer.getId());
+        if (foundPlayer == null) {
+            synchronized (activeAudioPlayers) {
```
muff1nman: We need to be very careful about synchronized as it can cause deadlock and performance issues. In this case, I think it would be more fitting to either have two synchronized (one added before line 50), or perhaps just make the entire method synchronized. If this synchronized block takes time to run, my fear is that a bunch of threads could get piled up right after checking for a null foundPlayer and end up doing extraneous work that the previous thread already did.

libresonic-main/src/main/java/org/libresonic/player/service/JukeboxJavaService.java
```diff
@@ -0,0 +1,299 @@
...
+        User user = securityService.getUserByName(libresonicPlayer.getUsername());
+        if (!user.isJukeboxRole()) {
+            log.warn("{} is not authorized for jukebox playback.",user.getUsername());
+            return;
+        }
+
+        log.debug("PlayQueue.Status is {}",libresonicPlayer.getPlayQueue().getStatus());
+        if (libresonicPlayer.getPlayQueue().getStatus() == PlayQueue.Status.PLAYING) {
+            MediaFile currentFileInPlayQueue;
+            synchronized (libresonicPlayer.getPlayQueue()) {
```
muff1nman: What is the need for synchronized here?

libresonic-main/src/main/java/org/libresonic/player/domain/Player.java
```diff
@@ -44,6 +44,7 @@
     private Date lastSeen;
     private TranscodeScheme transcodeScheme = TranscodeScheme.OFF;
     private PlayQueue playQueue;
+    private String javaJukeboxMixer;
```
muff1nman: What is this for? Just not quite understanding..

libresonic-main/src/main/java/org/libresonic/player/domain/PlayQueue.java
```diff
@@ -49,6 +51,29 @@
...
+            return null;
+        }
+    }
+
+    @Override
+    public int getSize() {
+        return size();
+    }
+    /* End : implementation of com.github.biconou.AudioPlayer.PlayList */
+
```
muff1nman: This doesn't seem like a good abstraction as `PlayQueue` seems like a separate entity from `Playlist`. The telling hint here, is that `Playlist` only deals with `java.io.File` where `PlayQueue` deals with `org.libresonic.player.domain.MediaFile`

libresonic-main/src/main/java/org/libresonic/player/service/JukeboxJavaService.java
```diff
@@ -0,0 +1,299 @@
...
+    private final static String DEFAULT_MIXER_ENTRY_KEY = "_default";
+
+    /**
+     * Finds the corresponding active audio player for a given libresonic player.
+     * The JukeboxJavaService references all active audio players in a map indexed by libresonic player id.
+     *
+     * @param libresonicPlayer a given libresonic player.
+     * @return the corresponding active audio player of null if none exists.
+     */
+    private com.github.biconou.AudioPlayer.api.Player retrieveAudioPlayerForLibresonicPlayer(Player libresonicPlayer) {
```
muff1nman: Sorry if I'm being naive here, but it seems like it would greatly simplify things if we could assume there can only be one jukebox player active at a time. This doesn't necessarily just apply to this method, but overall I think it would also simplify the jukebox interface. In the context of this method, it would make it so that there is only three states across the whole libresonic instance (not just specific to the Player): a single active `com.github.biconou.AudioPlayer.api.Player` libresonic instance, a single active legacy jukebox, or no active jukeboxes at all.

libresonic-main/src/main/java/org/libresonic/player/ajax/PlayQueueInfo.java
```diff
@@ -31,6 +31,7 @@
 public class PlayQueueInfo {
 
     private final List<Entry> entries;
+    private final int index;
```
muff1nman: I guess I just dont see where this is used. I think you said before its used in the frontend in a callback named nowPlayingInfos. Can you point out the file and line number?

libresonic-main/src/main/webapp/WEB-INF/jsp/playQueue.jsp
```diff
@@ -405,7 +410,13 @@
         });
     }
 
+    function isJavaJukeboxPresent() {
+        return $("#javaJukeboxPlayerControlBarContainer").length==1;
+    }
+
+
     function playQueueCallback(playQueue) {
+console.log("playQueueCallback");
```
muff1nman: Indentation.

libresonic-main/src/main/java/org/libresonic/player/service/JukeboxJavaService.java
biconou: Here are some explainations. I'll try to be as clear as possible but it's not easy.

A `com.github.biconou.AudioPlayer.api.Player` instance is the objects that really plays the music (send data to the audio line). It has the same role as `org.libresonic.player.service.jukebox.AudioPlayer` in the legacy jukebox.
An active `com.github.biconou.AudioPlayer.api.Player` is an instance associated to a `org.libresonic.player.domain.Player` and is created once the user first starts to play a playqueue for that `org.libresonic.player.domain.Player`.
Indeed there are possibly multiple active AudioPlayer at a time and they are all referenced in an association map help by the JukeboxJavaService.
There are two main reasons why we must maintain multiple AudioPlayer alive : 

1. The new jukebox is designed to let the user to choose a sound card to play to. It's a key feature of this new jukebox. Thus we can have multiple jukebox players playing on different sound cards. 
2. The AudioPlayer is part of a library I am working on independently from Libresonic. The AudioPlayer is designed to be standalone. It works with its own playlist. This playlist must be synchronized with the Libresonic playQueue and thus it must be associated with the Libresonic player holding the playQueue.

Details about 1.
The ability to choose the desired soundcard is very important as for JavaSound API, the default sound card is not necessarily the one you want. 
Being able to play at a time on multiple soundcards is less important but : 
- Suppose you want to control the jukebox on a same soundcard from both the web UI and a mobile app. These two controller devices started their own playqueue and it is important not to erase the AudioPlayer playlist when swapping from one controller to another. The new jukebox is designed so that, if these two devices address the same soundcard, if one starts, the other stops but you can swap from a controller device to another.
- In the future, I'll work on a remote jukebox player that will allow you to play music on a remote sound card from the Libresonic server. This is another reason why I choose to keep multiple AudioPlayers alive.

Details about 2.
I must confess that, for the moment, the relationship between objects that are involved in the jukebox process is messy. I have another organization idea in mind but I just wanted to propose a first version of my new jukebox without having to change too much things.
The guidelines for the aimed new code organization are : 

- The JavaJukeboxService annoys me. It sucks.
- The PlayQueue is the key object that should control a player. Thus the PlayQueueService (as it got methods like play/stop/add etc .... ) should directly control the AudioPlayer.
- The duplication/synchronization of PlayQueue/PlayList is a pain. I'd like to reorganize my AudioPlayer Library in a way that the AudioPlayer will directly be able to use the Libresonic PlayQueue as a PlayList.
- The new Jukebox should be always controlled through Rest API (at the moment, that's only the case with the mobile app).

libresonic-main/src/main/java/org/libresonic/player/domain/Player.java
biconou: It aims to memorize the chosen soundcard for a particular Libresonic player.
Mixer is for the JavaSound API a synonym to soundcard .

libresonic-main/src/main/java/org/libresonic/player/domain/PlayQueue.java
biconou: Did I answer in the previous huge comment ?

libresonic-main/src/main/java/org/libresonic/player/ajax/PlayQueueInfo.java
biconou: I have a doubt. May be it's no more used. I'll try to remove it.

libresonic-main/src/main/webapp/WEB-INF/jsp/playQueue.jsp
biconou: It was for debug purpose. I'll remove.

libresonic-main/src/main/java/org/libresonic/player/service/JukeboxJavaService.java
biconou: To prevent possibly parallel ajax request to create and AudioPlayer twice. Note that the synchroniezed block is only invoqued in case the foundPlayer was null, says the first time a jukebox player actives this service.

biconou: Note that I had issues without this synchronization.

libresonic-main/src/main/java/org/libresonic/player/service/JukeboxJavaService.java
biconou: Not sure. I'll remove it.

libresonic-main/src/main/java/org/libresonic/player/service/JukeboxJavaService.java
biconou: Not sure. I'll remove it.

biconou:
@muff1nman 

> In light of that, does it make more sense to move the "switch" between the jukebox technologies to the global settings level, rather than at the player level?

I'd prefer to wait till some real jukebox users test it and give their opinions.

libresonic-main/src/main/java/org/libresonic/player/service/JukeboxJavaService.java
muff1nman: So can we make the whole method synchronized?

libresonic-main/src/main/java/org/libresonic/player/domain/Player.java
muff1nman: Makes sense.

libresonic-main/src/main/java/org/libresonic/player/service/JukeboxJavaService.java
biconou: Will it not put a lock on the entire class ?

libresonic-main/src/main/java/org/libresonic/player/domain/PlayQueue.java
muff1nman: Yes, but I reallly think we still need to separate this out.

biconou: I need to think about it.

biconou:
This PR is now up to date with html5 player.

Issue #451, opened 2017-06-03
DLNA: more options for browsing (album/artist/genre)


akpdw:
The current implementation of the DLNA server only allows you to browse by folder and playlist.  This would allow for browsing the music library by album, artist, and genre, plus access to recently added albums.

This also adds limited support for search, so a client could search by album, artist, or song.

akpdw:
I don't see that I can assign myself to this issue, but I have a pull request more or less ready.

PR #452, opened 2017-06-03
451 dlna browse

akpdw:
Fixes Libresonic/libresonic#451

This replaces the existing FolderBasedContentDirectory with a new DispatchingContentDirectory.  The new system then dispatches requests to subclasses of UpnpContentProcessor, which then handle the requests as appropriate.

This also adds a new searchSongsByName method to MediaFileDao, which is used for the search by song functionality.

### DIFF
--- a/libresonic-main/src/main/java/org/libresonic/player/service/upnp/ArtistUpnpProcessor.java
+++ b/libresonic-main/src/main/java/org/libresonic/player/service/upnp/ArtistUpnpProcessor.java
@@ -1,0 +1,131 @@
+/*
+ This file is part of Libresonic.
+
+ Libresonic is free software: you can redistribute it and/or modify
+ it under the terms of the GNU General Public License as published by
+ the Free Software Foundation, either version 3 of the License, or
+ (at your option) any later version.
+
+ Libresonic is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU General Public License for more details.
+
+ You should have received a copy of the GNU General Public License
+ along with Libresonic.  If not, see <http://www.gnu.org/licenses/>.
+
+ Copyright 2017 (C) Libresonic Authors
+ Based upon Subsonic, Copyright 2009 (C) Sindre Mehus
+ */
+package org.libresonic.player.service.upnp;
+
+import org.fourthline.cling.support.contentdirectory.ContentDirectoryException;
+import org.fourthline.cling.support.model.BrowseResult;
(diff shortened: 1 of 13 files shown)


libresonic-main/src/main/java/org/libresonic/player/service/upnp/DispatchingContentDirectory.java
```diff
@@ -0,0 +1,281 @@
...
+      returnValue = getArtistProcessor().searchByName(titleSearch, firstResult, maxResults, orderBy);
+    } else if ("object.item.audioItem".equalsIgnoreCase(upnpClass)) {
+      returnValue = getMediaFileProcessor().searchByName(titleSearch, firstResult, maxResults, orderBy);
+    } else if ("object.container.album.musicAlbum".equalsIgnoreCase(upnpClass)) {
+      returnValue = getAlbumProcessor().searchByName(titleSearch, firstResult, maxResults, orderBy);
+    }
+
+    return returnValue != null ? returnValue : super.search(containerId, searchCriteria, filter, firstResult, maxResults, orderBy);
+  }
+
```
muff1nman: Can we use the searchService here to unify the search implementation across the front end and the dlna stuff?

libresonic-main/src/main/java/org/libresonic/player/service/upnp/ArtistUpnpProcessor.java
```diff
@@ -0,0 +1,131 @@
...
+                                   long firstResult, long maxResults,
+                                   SortCriterion[] orderBy)
+    throws ContentDirectoryException {
+
+    DIDLContent didl = new DIDLContent();
+
+    List<Artist> allArtists = getAllItems();
+    List<Artist> matches = new ArrayList<Artist>();
+    Pattern searchPattern = Pattern.compile(name, Pattern.CASE_INSENSITIVE);
+    for (Artist a: allArtists) {
```
muff1nman: lol, Are you finding it useful to be able to write regex queries in your dlna client? I like the idea, but I feel that a better idea would be to allow custom lucene queries instead. I'm not against the regex idea, but my concern is that this would be impossible/cumbersome to scale up against huge libraries where lucene is more appropriate.

akpdw: Yeah, no problem.  I had more or less based my approach on mimicking the various controllers and being as non-invasive as possible.  But I can go through and use the SearchService instead.  

(And fix up the merge conflict; I thought I was fully updated, so was surprised to see it...)

akpdw:
Cleaning up the branch ended up looking more difficult than just pushing out a new branch/pull request, so I just did that.  https://github.com/Libresonic/libresonic/pull/463

Issue #448, opened 2017-05-31
Mediaelement.js Placeholder issue


muff1nman:
This issue is to track the larger progress on mediaelment.js migration effort. So far we have the basics in place from #425. The updated checklist of remaining items is as follows:

#### PR Merged
- [x] Liquibase changes reverted
- [x] New Liquibase migration for any new transcodings we want/need to use with mediaelement to at least get it working OR see if `video/flv` works
- [X] steps for reproducing the mediaelement sources (i.e. in case of upgrade)

#### Develop Merged to Stable
- [ ] ~Remove duplicate code between audio and video `CastPlayer`s~ EDIT: descoped
- [ ] ~Move `CastPlayer` javascript embedded in videoPlayer/playQueue.jsp to standalone js~ EDIT: descoped
- [ ] ~Chrome cast support (use mediaelement plugin)~ EDIT: descoped
- [ ] Full screen control OR use ME controls for video (although I'd like to see this one done as part of the PR, it might be dependent on some of these other tasks)
- [ ] Stylizing of ME to match themes (#457)
- [ ] Update external player to use ME
- [ ] Remove jwplayer sources

#### Future work
- [ ] Support no transcode in the case none is needed.
- [ ] Change format type to use accept header (allows proper negotiation with server without a statically set format)
- [ ] Research into the most optimal video formats to use and prioritize them.

muff1nman:
@Eusmilis I took a crack at the css to at least make it look somewhat like ME fits in. Let me know if it looks like trash or if you have any suggestions (I haven't done css in a while :)

muff1nman:
@Eusmilis any luck on trying out the builtin chromecast support for mediaelement?

Eusmilis:
I made some progress, but I didn't have much time lately.

I changed the security config to remove the need for a jwt token, but the chromecast plugin for MediaElement still doesn't work. When I switch to chromecast, it casts about 1 second and stops.

However, after fixing the source for `cast_sender.js` the video `CastPlayer` works fine.

muff1nman:
Okay so perhaps it's not worth the effort to move the casting implementation? Also feel free to submit a PR to fix the source for cast_sender.js.

Eusmilis:
Since making the plugin work would require a lot of work and wouldn't be enough to get rid of the current UI (Using the MediaElement video player, the progress bar only show what has been buffered), I don't think it is worth it for now.

I still have to add a security token to the stream url before submitting a PR.

[...]

PR #457, opened 2017-06-08
MediaElement ui for default theme

muff1nman:
First crack at stylizing mediaelement. #448


muff1nman:
Some screenshots:
screenshot at 2017-06-07 19-37-37
screenshot at 2017-06-07 19-38-31
screenshot at 2017-06-07 19-39-22

I basically created a "light" theme for media element to use for the default. Where it was obvious I changed the theme back to the original media element theme for the darker themes.

jooola:
Why not a transparent player ? so it fits always the background colors ?

muff1nman:
That would be nice, feel free to submit a new PR with your idea.

Eusmilis:
I agree with @jooola, but we need to be careful with a transparent player, it could be hard to see on the lighter themes.
We could try to make a black version of `mejs-controls.svg` and set it in the theme's css if needed.

SuperNerd-James:
To be honest, I don't like the new player. The timer doesn't count up, so there's no way of knowing where you are exactly in a track (I use it for DJing), the previous and next track buttons are too small. Fast Forward and Rewind only work once you click the player first, and then use the left and right arrows.

I prefer the UI layout of Subsonic and Madsonic, if a reskin like that could work? As a separate skin so we have customise options?

biconou:
@SuperNerd-James I did not work on that feature so I can be mistaken but MediaElements, the library used here, can certainly permit some extended features. This was a first attempt to get rid of flash and you could suggest some improvements by creating new issues. 

DJing with Libresonic is not the best way. Did you try some real DJing softs ? Personally I use https://www.mixxx.org/ and that's really great.

muff1nman:
It might be good to take some time to think on how we'd like the interface to look with regards to the entire bottom playQueue frame. Thinking about this today, one way I could see doing it is:

- Move the entire play queue list that scrolls up from the bottom on mouse over to the NowPlaying view. The NowPlaying view currently seems pretty useless.
- Move controls related to the current play queue to the NowPlaying view.
- Expand the player to along the bottom and perhaps center it.
- Move the arrows into the media element player so that they're on either side of the play/pause button.
- Axe the entire right frame that displays the current playing information. I think its pretty useless except for the current playing album art. Perhaps we can display the album art in the bottom frame?

SuperNerd-James:
@muff1nman I agree with you on several points:

1. The current playing info is useless, if we get rid of it we'd have more space.

2. The PlayQueue does need work. I like it, but the fact it pops up intrusively needs changing.

3 The Now Playing view, is this the view that displays current album info? I quite like that.

I plan on becoming a possible developer here with you guys, but at the mo I'm more a .NET guy, I know bits of Java, and quite a lot of PHP and CSS etc, so once I'm fully familiar with the codebase and how the dev side of GitHub works, I can help submit patches.

@biconou Saying Libresonic is no good at DJ really depends on what you interpret as DJing. I don't mean the teen-kid-scratchy-disc type of DJing, I mean the running from a laptop at parties and clubs off playlists kind, and Libresonic is actually very good at it with a centralised server over the internet, it'll be even more once it's finished and I'm in it for the long haul :)

PR #461, opened 2017-06-09
Subsonic api 1.15.0

muff1nman:
Reverts #329 as I think it was causing issues. Updates libresonic to implement subsonic api 1.15.0 with the following exceptions:

getChatMessages returns 410 Gone
addChatMessage returns 410 Gone
getVideoInfo returns 501 Not implemented
getCaptions returns 501 Not implemented

`converted` parameter is not yet supported for /rest/stream
playlist is not returned on playlist creation

### DIFF
--- a/libresonic-main/src/main/java/org/libresonic/player/controller/JAXBWriter.java
+++ b/libresonic-main/src/main/java/org/libresonic/player/controller/JAXBWriter.java
@@ -26,12 +26,12 @@ import org.jdom.Attribute;
 import org.jdom.Document;
 import org.jdom.input.SAXBuilder;
 import org.libresonic.player.util.StringUtil;
-import org.libresonic.restapi.Error;
-import org.libresonic.restapi.ObjectFactory;
-import org.libresonic.restapi.Response;
-import org.libresonic.restapi.ResponseStatus;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.subsonic.restapi.Error;
+import org.subsonic.restapi.ObjectFactory;
+import org.subsonic.restapi.Response;
+import org.subsonic.restapi.ResponseStatus;
 
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
@@ -40,6 +40,7 @@ import javax.xml.bind.Marshaller;
 import javax.xml.datatype.DatatypeFactory;
 import javax.xml.datatype.XMLGregorianCalendar;
 
+import java.io.IOException;
 import java.io.InputStream;
 import java.io.StringWriter;
 import java.util.Date;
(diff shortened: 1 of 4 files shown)


libresonic-main/src/main/java/org/libresonic/player/controller/JAXBWriter.java
```diff
@@ -26,12 +26,12 @@
 import org.jdom.Document;
 import org.jdom.input.SAXBuilder;
 import org.libresonic.player.util.StringUtil;
-import org.libresonic.restapi.Error;
-import org.libresonic.restapi.ObjectFactory;
-import org.libresonic.restapi.Response;
-import org.libresonic.restapi.ResponseStatus;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.subsonic.restapi.Error;
```
biconou: Package name of generated files is back to org.subsonic. Is it due to the name space in the XSD file ? 
Did you decide to keep the XSD file consistent with this one ? http://www.subsonic.org/pages/inc/api/schema/subsonic-rest-api-1.15.0.xsd

libresonic-main/src/main/java/org/libresonic/player/controller/RESTController.java
```diff
@@ -2218,6 +2240,87 @@ public void setRating(HttpServletRequest request, HttpServletResponse response)
...
+    public ResponseEntity<String> getVideoInfo() throws Exception {
+        return ResponseEntity.status(HttpStatus.SC_NOT_IMPLEMENTED).body(NOT_YET_IMPLEMENTED);
+    }
+
+    @RequestMapping(value = "/getCaptions", method = RequestMethod.GET)
+    public ResponseEntity<String> getCaptions() {
+        return ResponseEntity.status(HttpStatus.SC_NOT_IMPLEMENTED).body(NOT_YET_IMPLEMENTED);
+    }
+
+    @RequestMapping(value = "/startScan", method = {RequestMethod.PUT})
```
biconou: Is it really a PUT ?

libresonic-main/src/main/java/org/libresonic/player/controller/JAXBWriter.java
muff1nman: Yeah, my thought was that we should keep it consistent if we are going to have a subsonic-compatible api. I havent done it yet, but my thought was to put all libresonic specific api stuff somewhere else. Do you think its a bad idea?

libresonic-main/src/main/java/org/libresonic/player/controller/RESTController.java
muff1nman: Good catch, this should be a POST.

libresonic-main/src/main/java/org/libresonic/player/controller/JAXBWriter.java
biconou: No I like it. We could consider that Libresonic is not subsonic but is subsonic api complient. Reach idea.

libresonic-main/src/main/java/org/libresonic/player/controller/RESTController.java
biconou: I meant I am afraid it's a get as it is a subsonic api method and everything is get in this api (not a properly designed true rest app).

muff1nman: Really where do you see that it's a GET? The subsonic  API docs don't really say one way or the other..

biconou: You're right. Does it mean that any http method can be used ?

muff1nman: I'm not sure, I was just applying my best judgment on GET vs POST based on the semantics of the method.

muff1nman: Might be best to keep with what the other methods do: both GET and POST:
```
    @RequestMapping(value = "/getTopSongs", method = {RequestMethod.GET, RequestMethod.POST})
```

PR #463, opened 2017-06-10
Fixes Libresonic/libresonic#451

akpdw:
This replaces the existing FolderBasedContentDirectory with a new
DispatchingContentDirectory.  The new system then dispatches requests to
subclasses of UpnpContentProcessor, which then handle the requests as
appropriate.

This version also implements a new searchByName method in SearchService,
which is used to provide artist, album, and track search functionality
for the upnp service.

Guidelines for Contributing
---------------------------

Libresonic development is a community project, and contributions are welcomed. Here are a few guidelines you should follow before submitting:

  1.  **License Acceptance** All contributions must be licensed as [GNU GPLv3](https://github.com/Libresonic/libresonic/blob/develop/LICENSE.txt) to be accepted. Use [`git commit --signoff`](https://jk.gs/git-commit.html) to acknowledge this.
  2.  **No Breakage** New features or changes to existing ones must not degrade the user experience. This means do not introduce bugs, remove functionality, or make large changes to existing themes/UI without prior discussion in an Issue.
  3.  **Coding standards** Language best-practices should be followed, comment generously, and avoid "clever" algorithms. Refactoring existing messes is great, but watch out for breakage.
  4.  **Be bold!** Without contributions, this project will vanish. If you just want to help out, try [submiting a patch](https://github.com/Libresonic/libresonic/issues?q=is%3Aissue+is%3Aopen+label%3Apatches-welcome) for an open Issue.
  5.  **Stay relevant** Issues or commentary that is off-topic or tangential to Libresonic development is subject to moderation. Questions should be focused on improving documentation to solve a problem. Visit [Reddit](https://www.reddit.com/r/libresonic) or [IRC](http://webchat.freenode.net?channels=%23libresonic) for community discussion.

### DIFF
--- a/libresonic-main/src/main/java/org/libresonic/player/service/upnp/AlbumUpnpProcessor.java
+++ b/libresonic-main/src/main/java/org/libresonic/player/service/upnp/AlbumUpnpProcessor.java
@@ -18,8 +18,10 @@
   Based upon Subsonic, Copyright 2009 (C) Sindre Mehus
 */
 package org.libresonic.player.service.upnp;
+import org.fourthline.cling.support.model.BrowseResult;
 import org.fourthline.cling.support.model.DIDLContent;
 import org.fourthline.cling.support.model.PersonWithRole;
+import org.fourthline.cling.support.model.SortCriterion;
 import org.fourthline.cling.support.model.container.Container;
 import org.fourthline.cling.support.model.container.MusicAlbum;
 import org.libresonic.player.dao.AlbumDao;
@@ -53,6 +55,20 @@ public class AlbumUpnpProcessor extends UpnpContentProcessor <Album, MediaFile> 
         setRootTitle("Albums");
     }
 
+    /**
+     * Browses the top-level content of a type.
+     */
+    public BrowseResult browseRoot(String filter, long firstResult, long maxResults, SortCriterion[] orderBy) throws Exception {
+        DIDLContent didl = new DIDLContent();
+
+        List<MusicFolder> allFolders = getDispatchingContentDirectory().getSettingsService().getAllMusicFolders();
+        List<Album> selectedItems = getAlbumDao().getAlphabeticalAlbums((int) firstResult, (int) maxResults, false, true, allFolders);
+        for (Album item : selectedItems) {
+            addItem(didl, item);
+        }
+
+        return createBrowseResult(didl, (int) didl.getCount(), getAllItemsSize());
+    }
     public Container createContainer(Album album) throws Exception {
         MusicAlbum container = new MusicAlbum();
 
@@ -74,7 +90,7 @@ public class AlbumUpnpProcessor extends UpnpContentProcessor <Album, MediaFile> 
 
     public List<Album> getAllItems() {
         List<MusicFolder> allFolders = getDispatchingContentDirectory().getSettingsService().getAllMusicFolders();
-        return getAlbumDao().getAlphabetialAlbums(0, 0, false, allFolders);
+        return getAlbumDao().getAlphabeticalAlbums(0, Integer.MAX_VALUE, false, true, allFolders);
     }
 
     public Album getItemById(String id) throws Exception {
@@ -109,6 +125,13 @@ public class AlbumUpnpProcessor extends UpnpContentProcessor <Album, MediaFile> 
         }
         return allFiles;
     }
+
+    @Override
+    public int getAllItemsSize() throws Exception {
+        List<MusicFolder> allFolders = getDispatchingContentDirectory().getSettingsService().getAllMusicFolders();
+        return getAlbumDao().getAlbumCount(allFolders);
+    }
+
 
     public void addChild(DIDLContent didl, MediaFile child) throws Exception {
         didl.addItem(getDispatcher().getMediaFileProcessor().createItem(child));
--- a/libresonic-main/src/main/java/org/libresonic/player/service/upnp/AlbumUpnpProcessor.java
+++ b/libresonic-main/src/main/java/org/libresonic/player/service/upnp/AlbumUpnpProcessor.java
@@ -1,0 +1,136 @@
+/*
+  This file is part of Libresonic.
+
+  Libresonic is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  Libresonic is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with Libresonic.  If not, see <http://www.gnu.org/licenses/>.
+
+  Copyright 2017 (C) Libresonic Authors
+  Based upon Subsonic, Copyright 2009 (C) Sindre Mehus
+*/
+package org.libresonic.player.service.upnp;
+import org.fourthline.cling.support.model.DIDLContent;
+import org.fourthline.cling.support.model.PersonWithRole;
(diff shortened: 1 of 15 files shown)


libresonic-main/src/main/java/org/libresonic/player/service/upnp/DispatchingContentDirectory.java
```diff
@@ -0,0 +1,292 @@
...
+import java.util.Arrays;
+
+/**
+ * @author Allen Petersen
+ * @author Sindre Mehus
+ * @version $Id$
+ */
+public class DispatchingContentDirectory extends LibresonicContentDirectory {
+
+    public static final Logger LOG = LoggerFactory.getLogger(FolderBasedContentDirectory.class);
```
muff1nman: Logger class should be `DispatchingContentDirectory`

libresonic-main/src/main/java/org/libresonic/player/service/upnp/DispatchingContentDirectory.java
```diff
@@ -0,0 +1,292 @@
...
+            String[] splitId = objectId.split(SEPARATOR);
+            String browseRoot = splitId[0];
+            String itemId = splitId.length == 1 ? null : splitId[1];
+
+            UpnpContentProcessor processor = findProcessor(browseRoot);
+            if (processor == null) {
+                // if it's null then assume it's a file, and that the id
+                // is all that's there.
+                itemId = browseRoot;
+                browseRoot = "folder";
```
muff1nman: Why is this needed? Looks like browseRoot isnt used after this point.

libresonic-main/src/main/java/org/libresonic/player/service/upnp/AlbumUpnpProcessor.java
```diff
@@ -0,0 +1,136 @@
...
+        // TODO: correct artist?
+        if (album.getArtist() != null) {
+            container.setArtists(getAlbumArtists(album.getArtist()));
+        }
+        return container;
+    }
+
+    public List<Album> getAllItems() {
+        List<MusicFolder> allFolders = getDispatchingContentDirectory().getSettingsService().getAllMusicFolders();
+        return getAlbumDao().getAlphabetialAlbums(0, 0, false, allFolders);
```
muff1nman: Album browsing didnt work for me and I think its because this only ever returns zero albums.

libresonic-main/src/main/java/org/libresonic/player/service/upnp/ArtistUpnpProcessor.java
```diff
@@ -0,0 +1,99 @@
...
+    public  List<Album> getChildren(Artist artist) {
+        List<MusicFolder> allFolders = getDispatcher().getSettingsService().getAllMusicFolders();
+        List<Album> allAlbums = getAlbumProcessor().getAlbumDao().getAlbumsForArtist(artist.getName(), allFolders);
+        if (allAlbums.size() > 1) {
+            Album viewAll = new Album();
+            viewAll.setName("- All Albums -");
+            viewAll.setId(-1);
+            viewAll.setComment(AlbumUpnpProcessor.ALL_BY_ARTIST + "_" + artist.getId());
+            allAlbums.add(0, viewAll);
+        }
```
muff1nman: What is the point of this? Wouldn't someone just go back to all artists if they wanted to. Seems strange to add custom navigation.

libresonic-main/src/main/java/org/libresonic/player/service/upnp/RecentAlbumUpnpProcessor.java
```diff
@@ -0,0 +1,50 @@
...
+    public List<Album> getAllItems() {
+        List<MusicFolder> allFolders = getDispatchingContentDirectory().getSettingsService().getAllMusicFolders();
+        List<Album> recentAlbums = getAlbumDao().getNewestAlbums(0, RECENT_COUNT, allFolders);
+        if (recentAlbums.size() > 1) {
+            Album viewAll = new Album();
+            viewAll.setName("- All Albums -");
+            viewAll.setId(-1);
+            viewAll.setComment(AlbumUpnpProcessor.ALL_RECENT);
+            recentAlbums.add(0, viewAll);
+        }
```
muff1nman: Same as in the artist processor, why do we need this?

libresonic-main/src/main/java/org/libresonic/player/service/upnp/ArtistUpnpProcessor.java
akpdw: Browsing by artist goes (list all artists) - (list albums by selected artist) - (list songs on selected album).  The '  - All Albums - ' option allows for the third step to be (list all songs on all albums by that artist).

I added a comment to the code to try to make it clearer.

libresonic-main/src/main/java/org/libresonic/player/service/upnp/RecentAlbumUpnpProcessor.java
akpdw: Same as above, in this case allowing you to select all songs on all recent albums.

libresonic-main/src/main/java/org/libresonic/player/service/upnp/AlbumUpnpProcessor.java
akpdw: Huh.  I think that album list is populated from ID3 tags?  My music collection loaded in, if not all, then at least most of my albums, but I didn't check all that closely.

akpdw:
I also added in the broadcast search capabilities change.  The two clients I've been using apparently don't pay attention to that at all.  DLNA is such a wonderfully strictly-supported spec. :)

libresonic-main/src/main/java/org/libresonic/player/service/upnp/ArtistUpnpProcessor.java
muff1nman: Sure makes sense. But it seems weird we are adding it as a `org.libresonic.player.domain.Album`. Really we should be adding this at the `org.fourthline.cling.support.model.container.Container` level which would also probably fix the display to look like a folder on upnp clients.

libresonic-main/src/main/java/org/libresonic/player/service/upnp/AlbumUpnpProcessor.java
muff1nman: Okay I think I figured out the issue. For hsql, it returns all records in the case that you send limit 0 which is weird behavior possibly related to https://sourceforge.net/p/hsqldb/bugs/1100/. However, being on postgres this is actually returning zero rows for me. The naive fix here is to change / add a DAO level function to return all albums in alpha order. However, the better change IMO would be to utilize the pagination options that were passed in at `org.libresonic.player.service.upnp.UpnpContentProcessor#browseRoot` rather than doing the in memory filtering as seen at `org/libresonic/player/service/upnp/UpnpContentProcessor.java:81`

libresonic-main/src/main/java/org/libresonic/player/service/upnp/ArtistUpnpProcessor.java
akpdw: It gets added as a org.libresonic.player.domain.Album to go along with the actual Albums that the Artist has.  Those then are passed on to addChild(), which in turn calls getAlbumProcessor().createContainer(album), and adds it to the response as a  org.fourthline.cling.support.model.container.MusicAlbum.  So it does get added as a Container eventually.

You could argue that since this isn't a proper MusicAlbum that it (both the all by artist and all recent albums) is really more of a playlist, and should be represented by a org.fourthline.cling.support.model.container.PlaylistContainer instead.  I don't know if that would gain anything on a practical level, but it could be handled that way instead.

muff1nman:
Good work here, merged this into the develop branch. Still need to test since the album changes but any further tweaks can be done in the develop branch.

PR #473, opened 2017-06-24
First step in javascript code refactoring.

biconou:
This PR is for review and discussion. 
It brings nothing new but it is a first step of a way to write a better js code. This is important to be able to turn progressively to a modern UI technology. 

This first step concentrates on the playQueue view that has a lot of js logic. 
I have extracted all js code in a separate playQueue.js file. 
I have created a javascript namespace called libresonic.playQueue with both private and plublic members. 

I use two transport objects to transfer datas from the jsp to javascript : 

- model : a javascript transposition of the Spring MVC model. 
- resources : a javascript object containing the needed i18n resources keeping the namespacing logic.

Review is required especially from @muff1nman.
Contributions from @fxthomas and @Eusmilis would be appreciated.

### DIFF
--- a/libresonic-main/src/main/webapp/WEB-INF/jsp/playQueue.jsp
+++ b/libresonic-main/src/main/webapp/WEB-INF/jsp/playQueue.jsp
@@ -12,6 +12,40 @@
     <script type="text/javascript" src="<c:url value="/dwr/util.js"/>"></script>
     <script type="text/javascript" src="<c:url value="/script/mediaelement/mediaelement-and-player.min.js"/>"></script>
     <%@ include file="playQueueCast.jsp" %>
+    <script type="text/javascript" src="<c:url value="/script/playQueue/playQueue.js"/>"></script>
+    <script type="text/javascript" language="javascript">
+        $(document).ready(function() {
+            var model = {
+                player : {
+                    web : ${model.player.web}
+                },
+                partyMode : ${model.partyMode},
+                notify : ${model.notify},
+                autoHide : ${model.autoHide},
+            };
+            var resources = {
+                common : {
+                    cancel : '<fmt:message key="common.cancel"/>'
+                },
+                playlist : {
+                    confirmclear : '<fmt:message key="playlist.confirmclear"/>',
+                    toast : {
+                        saveplayqueue : '<fmt:message key="playlist.toast.saveplayqueue"/>',
+                        saveasplaylist : '<fmt:message key="playlist.toast.saveasplaylist"/>',
+                        appendtoplaylist : '<fmt:message key="playlist.toast.appendtoplaylist"/>'
+                    },
+                    repeat_radio : '<fmt:message key="playlist.repeat_radio"/>',
+                    repeat_on : '<fmt:message key="playlist.repeat_on"/>',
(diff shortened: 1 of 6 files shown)


libresonic-main/src/main/webapp/WEB-INF/jsp/playQueue.jsp
```diff
@@ -12,6 +12,40 @@
...
+                    repeat_off : '<fmt:message key="playlist.repeat_off"/>'
+                },
+                playlist2 : {
+                    songs : '<fmt:message key="playlist2.songs"/>'
+                }
+            };
+            libresonic.playQueue.init(model,resources);
+        });
+    </script>
+
```
muff1nman: I like the idea here. It segregates the model/jsp data out and opens the door to move this to a AJAX call if so desired in the future.

libresonic-main/src/main/webapp/script/playQueue/playQueue.js
```diff
@@ -0,0 +1,675 @@
...
+        $("#moreActions").prop("selectedIndex", 0);
+    }, // actionSelected
+    onPlay : function(id) {
+        if (_debug) {
+          console.debug("onPlay");
+        }
+        playQueueService.play(id, playQueueCallback);
+    }
+  }
+})();
```
muff1nman: This module format looks pretty good to me with one suggestion. Should we move all of the globals to be injected into the module rather than just randomly accessing them? Like `playQueueService` and `playlistService`. It might improve clarity on what this module depends on.

biconou: Sure. That's a good idea. I'll try it as soon as I can since I will have few time these days.

biconou:
I definitely think it's worth taking some time to find the good module solution. 
I like the idea of investigating in the ES6 way.

biconou:
What about require.js ?

libresonic-main/src/main/webapp/WEB-INF/jsp/playQueue.jsp
```diff
@@ -711,34 +120,34 @@
...
-                        $("#jukeboxVolume").on("slidestop", onJukeboxVolumeChanged);
+                        $("#jukeboxVolume").on("slidestop", libresonic.playQueue.onJukeboxVolumeChanged);
                     </script>
                 </td>
             </c:if>
 
             <c:if test="${model.player.web}">
                 <td><span class="header">
-                    <img src="<spring:theme code="backImage"/>" alt="" onclick="onPrevious()" style="cursor:pointer"></span>
+                    <img src="<spring:theme code="backImage"/>" alt="" onclick="libresonic.playQueue.onPrevious()" style="cursor:pointer"></span>
```
fxthomas: While we're cleaning up things, why not move all these `onclick` handlers to something that binds to `click` events in the JavaScript module, in order to avoid mixing presentation and code?

biconou: @fxthomas It's planned. But this is a work in progress and the next step is to explore the best module syntax solution as mentionned by @muff1nman

PR #474, opened 2017-06-24
Libresonic rest api

muff1nman:
Still a work in progress, but just wanted to open a PR for visibility.

TODO:
- [ ] Use `@RequestParam` instead of `request.getParameter`
- [ ] Add proper documentation on swagger api output
- [ ] Move generated swagger json to something other than /v2/api-docs
- [ ] Perhaps generate swagger ahead of time to avoid having springfox dependencies in the war?
- [ ] Address authentication


muff1nman:
I'll update this PR later with details on everything I did here.

biconou:
What a work ! 
I don't got much time at the moment but I will be interested in reviewing this later. 
SpringFox seems to be something to generate a swagger documentation ? 
Is the idea behind the code first approach ?

libresonic-main/src/main/java/org/libresonic/player/annotation/LibresonicRESTApi.java
```diff
@@ -0,0 +1,4 @@
+package org.libresonic.player.annotation;
+
+public @interface LibresonicRESTApi {
```
biconou: What is this annotation for ?

libresonic-main/src/main/java/org/libresonic/player/controller/AvatarController.java
```diff
@@ -60,17 +59,16 @@ public long getLastModified(HttpServletRequest request) {
...
         if (avatar == null) {
             response.sendError(HttpServletResponse.SC_NOT_FOUND);
-            return null;
+            return;
         }
 
         response.setContentType(avatar.getMimeType());
         response.getOutputStream().write(avatar.getData());
-        return null;
     }
```
biconou: You changed a lot of controllers this way. Is is just a cleanup because the return type ModelAndView is useless ?

muff1nman: Correct.

libresonic-main/src/main/java/org/libresonic/player/annotation/LibresonicRESTApi.java
muff1nman: Its for the springfox docket so that it only scans this class

muff1nman:
Also fyi, I am in the process of adding in proper documentation using `@ApiOperation` and `@ApiResponse` and also adding in `@RequestParam` rather than using `request.getParameter`

muff1nman:
To answer your question, yes the idea is a code first with the schema (swagger) being generated from the code. I'm not opposed to doing a schema first approach if it makes sense.

muff1nman:
Updated top comment with todos.

biconou:
Code first is good for our situation.

PR #476, opened 2017-07-08
CUE sheet support for single-track albums

inkognito13:
This is implementation of support for single file album with cue file. 

Path of song from single-file album is stored as 
```
<whole file path>:<seconds from begin of file where song starts>:<seconds from begin of file where song ends>
```

Also there are two new params in transcoding string: %ss and %to, which represents same begin and end seconds of song. So example string for ffmpeg will look like this: 
```
ffmpeg -ss %ss -i %s -to %to -map 0:0 -b:a %bk -v 0 -f mp3 -
```

Parsing of .cue file is done with [cuelib](https://github.com/carlwilson/cuelib). I was not able to found it no maven central, so used [jitpack](https://jitpack.io)

### DIFF
--- a/libresonic-main/pom.xml
+++ b/libresonic-main/pom.xml
@@ -462,6 +462,11 @@
             <artifactId>jackson-databind</artifactId>
             <version>2.8.7</version>
         </dependency>
+        <dependency>
+            <groupId>com.github.carlwilson</groupId>
+            <artifactId>cuelib</artifactId>
+            <version>master-SNAPSHOT</version>
+        </dependency>
     </dependencies>
 
     <build>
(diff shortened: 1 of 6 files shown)


korden32:
I'm very new in Java and can't exactly tell you what's wrong, but it doesn't work for me.
CUE parsing is working, so I can see individual tracks in the list, but when I try to play it just throws

```
java.lang.SecurityException: Access denied to file Taikatalvi
        at org.libresonic.player.service.MediaFileService.getMediaFile(MediaFileService.java:143) ~[
classes!/:6.3-SNAPSHOT]
        at org.libresonic.player.controller.StreamController.getSingleFile(StreamController.java:276
) ~[classes!/:6.3-SNAPSHOT]
        at org.libresonic.player.controller.StreamController.handleRequest(StreamController.java:130
) ~[classes!/:6.3-SNAPSHOT]
...
```

jooola:
@inkognito13 Hello, would you like to bring you're PR on the airsonic/airsonic repo too ?
And issue is currenlty assigned on this PR : airsonic/airsonic#495

Yetangitu:
FYI, I made a working implementation of CUE support for Airsonic based on this PR: https://github.com/airsonic/airsonic/pull/856

It contains fixes and improvements which might also be of interest to this project.

